Issue #29, opened 2026-06-23
[Bug]: Huge memory usage ( oom activated ) while scanning


pinkfloydFR:
### What happened?

Hi,

After installing for the first time this application, i got a problem. The application is using huge memory ( 32 go ) while scanning my library ( nas folder )

Hope attachment can help you because nothing strange in the log during scanning.

[liste_fichiers_formatee.txt](https://github.com/user-attachments/files/29269332/liste_fichiers_formatee.txt)

My biggest files : 

<img width="520" height="298" alt="Image" src="https://github.com/user-attachments/assets/d7da9b00-b80a-421f-a0c4-263b3802c189" />

### Steps to reproduce

Manually launch a library refresh

### Area

File ingestion

### Version or commit

latest

### Environment

Docker under Ubuntu 24.04

### Relevant logs

```shell
sadly, nothing :(
```

EDIT : 
Oh, and FYI, don't think it's the cause, but for launching your apps, i needed to delete the command section for "printstash-api" ( IA helped me with this)

With the command in docker compose i got : 
```"[printstash-api] 2026-06-23T20:41:52.206453905Z /usr/local/bin/python: No module named alembic.__main__; 'alembic' is a package and cannot be directly executed"```  and the api docker crashed at start

Without the command section : all start without any error in log, docker, latest

```
services:
  printstash-api:
    image: ghcr.io/xiao-villamor/printstash-api:latest
    container_name: printstash-api
    logging: *default-logging
    environment:
      TZ: Europe/Paris
      VAULT_JWT_SECRET: "xxxx"
      VAULT_MAX_UPLOAD_MB: "512"
      VAULT_BACKUP_RETENTION_DAYS: "30"
      VAULT_ACCESS_TOKEN_EXPIRE_MINUTES: "60"
      VAULT_LOG_LEVEL: "INFO"
    volumes:
      - ./data/db:/data/db
      - ./data/thumbs:/data/thumbs
      - ./data/staging:/data/staging
      - ./data/backups:/data/backups
      - /mnt/docs/Impression_3D/06-STL:/mnt/docs/Impression_3D/06-STL
    expose:
      - "8000"
    restart: "no"
    networks:
      printstash:
        aliases:
          - api
      traefik:
```

xiao-villamor:
[...]

PR #31, opened 2026-06-24
docs: Alembic migration troubleshooting + un-stamped DB repair (#29)

xiao-villamor:
## What & why

A user in #29 hit `No module named alembic.__main__` at startup. Root cause: the shipped Compose command (`uv run alembic upgrade head`) is correct, but a hand-written `python -m alembic …` substitution isn't — Alembic ships no runnable `__main__`. They then "fixed" it by deleting the `command:` block, which lets the app bootstrap the schema via `create_all()` **without recording a migration version** — a latent `table already exists` failure on the next `alembic upgrade head`.

## Change (docs only)

Adds a **Troubleshooting migrations** section to the published upgrade guide and a matching note in repo-root `UPGRADE.md`:

- Correct command is `uv run alembic upgrade head` (console script) / `python -m alembic.config …` — never `python -m alembic`.
- Data-safe one-time repair for an un-stamped DB: `alembic stamp head` (writes only the version marker, touches no data), then upgrade normally.

No runtime, Dockerfile, or Compose changes. The entrypoint hardening (run migrations in the image + remove the Compose command + auto-stamp guard) is intentionally left for a separate, tested change.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -57,6 +57,19 @@ Then restart the backend and frontend development servers.
 - Always back up the SQLite file before running migrations.
 - Do not edit the SQLite file directly while the API container is running.
 
+## Troubleshooting Migrations
+
+- **`No module named alembic.__main__`** — run migrations with the `alembic`
+  console script (`docker compose run --rm api uv run alembic upgrade head`), not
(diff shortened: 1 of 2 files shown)


xiao-villamor:
Superseded by #33 (consolidated onto the 0.7.2 release branch).

PR #32, opened 2026-06-24
fix(web): keep collection context on delete and logo navigation

xiao-villamor:
## Problem

Two navigation papercuts kicked users out of the folder they were browsing:

1. **Deleting a model** returned to **All Models** instead of staying in the model's collection. Removing one model from `spoolers` dumped you at the root.
2. **The PrintStash logo** always linked to `/`, so going **Settings → logo** landed on All Models rather than the collection you came from (the user expected browser-back-like behavior).

## Fix

Collection selection lives in the URL as `?c=<path>` but is dropped when you leave the browser (Settings, a model, etc.). This remembers the last-browsed collection so we can restore it:

- New `src/lib/last-collection.ts` helper (`rememberLastCollection` / `readLastCollection` / `lastVaultHref`), backed by `localStorage`, best-effort (guards private mode).
- The model grid records the active folder whenever it changes.
- **Logo** (`top-bar.tsx`) now restores the last collection instead of hard-linking to `/`, recomputed on route change.
- **Post-delete** (`model-detail/index.tsx`) navigates to `/?c=<model.collection>` (or `/` if uncollected).

## Tests

- Unit tests for the helper (roundtrip, encoding of nested/spaced paths, clear-at-root).
- `tsc` clean; eslint clean on changed files (only pre-existing warnings remain).
- Existing `app-routes` e2e navigation spec still passes (6/6).

Note: from within the model browser the logo still effectively reflects the folder you arrived on; the reported case (returning from Settings/other pages) is what this restores.

### DIFF
--- a/frontend/src/components/model-detail/index.tsx
+++ b/frontend/src/components/model-detail/index.tsx
@@ -168,7 +168,11 @@ export function ModelDetail({ model: initialModel }: { model: ModelRead }) {
     try {
       await deleteModel(model.id);
       toast.success("Model deleted");
-      router.push("/");
+      // Return to the folder the model lived in, not the root — deleting one
+      // model shouldn't kick the user out of the collection they were browsing.
+      router.push(
+        model.collection ? `/?c=${encodeURIComponent(model.collection)}` : "/",
+      );
       router.refresh();
     } catch (e) {
       toast.error(e);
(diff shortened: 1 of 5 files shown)


xiao-villamor:
Superseded by #33 (consolidated onto the 0.7.2 release branch).

PR #51, opened 2026-07-12
feat(frontend): design-system consolidation & motion overhaul

jorgehermo9:
## What

A frontend-wide design-system and motion overhaul, executed from seven self-contained implementation plans (included under `docs/plans/`, written from a six-agent audit of the codebase at `57770ff`).

### Foundations
- **Motion tokens** (`--ease-out`, `--ease-in-out`, `--duration-press/fast/slow`) — every keyframe/utility now consumes them; app-wide `prefers-reduced-motion` policy (movement drops, opacity feedback stays); bouncy theme-icon curve replaced with the house crisp ease-out.
- **Token unification** — ~837 arbitrary `[var(--…)]` color classes converted to mapped Tailwind classes; all `blue-* dark:orange-*` primary hand-encodings replaced with `primary`/`accent`/`ring` tokens; new `primary-hover`, `primary-soft`, `success`, `warning` tokens; real radius scale (`rounded` now on-token); `z-dropdown`/`z-overlay` scale; `text-2xs`/`text-3xs` type ramp; fixes two undefined-variable bugs (`--on-primary`, `--surface-variant`).

### Primitives (new or hardened, in `frontend/src/components/ui/`)
- **ModalShell / Modal / Drawer** — portal, focus trap + restore, Escape, scroll lock, animated symmetric enter/exit (transition-based, interruptible). 8 bespoke overlays migrated; drawers no longer vanish instantly on close.
- **DropdownMenu + useComboboxNav** — origin-aware scale/fade from the trigger, outside-pointerdown dismiss, Escape-refocuses-trigger, arrow-key roving; listbox ARIA + arrow-key highlight for all 3 chip editors and 2 category pickers.
- **Button** (press feedback, `loading`, `xs`/`icon-sm`), **Input** (`aria-invalid`), **Checkbox** (forwardRef/disabled/tokens), **Badge** (`success`/`warning`), **Spinner**, **EmptyState**, **TabBar** (ARIA tablist, sliding indicator). Per-file `BTN_*`/`INPUT` constants now delegate to `buttonVariants`/`inputClasses`.

### Feel sweeps
- Zero `transition-all` (was 22); progress bars now GPU-composited `scaleX` with `linear` timing; press feedback on model cards and nav rows with transform in the transition list.
- `panel-in` entrance (150ms/4px) on tab panels, settings sections, grid↔list toggle, and empty states.

## How to review — flow-by-flow

The motion is deliberately subtle (crisp dashboard, not showy), so review by *interacting*, ideally with `main` running side by side. Ordered by how visible the difference is:

**No data needed**

1. **Upload modal** (`+` / Upload button): backdrop fades in and the panel scales 0.97→1 over ~200ms, and it **animates out** on close/Escape. Before: popped in and vanished in one frame. Also: focus is trapped inside (Tab cycles), and closing returns focus to the opener.
2. **Theme toggle** (sun/moon): the icon used to do a bouncy overshoot spin growing from 60% scale; now it settles crisply (rotate −30°, scale 0.9→1, no bounce).
3. **Dark mode — any primary button** (*static* diff): text on primary/destructive buttons is now dark-on-orange / dark-on-red per the `--*-foreground` tokens (was white-on-orange/red). Correct contrast pairing, matches the FAB.
4. **Press-and-hold any button** (Settings is full of them): compresses to 98% over 150ms and springs back on release. Before: no press feedback anywhere except the FAB/bottom nav.
5. **Profile & notification menus** (top-right): scale in from their trigger corner (transform-origin top-right) instead of appearing instantly; ArrowUp/Down/Home/End walk the items; Escape closes and refocuses the trigger; outside-click dismiss now also works for touch/pen (pointerdown).
6. **Mobile drawers** (narrow viewport → hamburger, or the bottom-nav "More" sheet): closing now slides/settles out symmetrically — before, the drawer keyframed in but was torn out instantly on close. Escape now closes them; while sliding in, tapping the backdrop reverses the motion mid-flight (transitions retarget; the old keyframes couldn't).
7. **Settings tabs**: the active underline **slides** between tabs (and retargets if you click quickly) and each section settles in with a 4px rise. Before: indicator blinked out/in, content hard-swapped. The strip is now a real ARIA tablist — focus it and use ArrowLeft/Right.
8. **Light theme — focus any input**: focus rings are always theme-correct (blue in light / orange in dark). Before, several inputs had hardcoded `focus:ring-orange-500`, orange even in light mode.

**After uploading a model or two**

9. **Model cards**: hover shows the same border/shadow feedback, but press-and-hold now compresses the card ~1%; `transition-all` is gone so only intended properties animate.
10. **Upload progress bars** (task popover, printer pages): fill at constant linear speed via GPU `transform: scaleX()` — before they eased a layout-triggering `width` at 500ms, visibly lurching.
11. **Model detail tabs** (Overview/Files/Revisions…): sliding indicator + panel settle, same treatment as Settings; switching tabs also resets panel scroll to top (intended).
12. **Grid ↔ list toggle**: the new layout fades/rises in instead of hard-cutting; emptying the grid with a filter makes "No models found" settle in instead of popping.
13. **Tag editing** (batch toolbar chip editor, model overview tags, upload modal tags): type a letter, ArrowUp/Down cycles a highlighted suggestion, Enter picks it (or creates the typed text with no highlight), Backspace on empty input still removes the last chip. The lists are proper `role="listbox"`/`option` comboboxes with `aria-activedescendant`. Category pickers animate from the trigger and lost their invisible full-screen click-catcher divs.

**Cross-cutting checks**

- **Reduced motion** (DevTools → Rendering → emulate `prefers-reduced-motion`): drawers/menus/panels degrade to opacity fades — movement drops, feedback stays. Before, only the card entrance was gated.
- **Slow-motion inspection** (DevTools → Animations panel at 10%): modal/menu curves are the strong house ease-out (`cubic-bezier(0.16, 1, 0.3, 1)`), durations 150–200ms, nothing over 300ms.
- **Keyboard-only pass**: every overlay (modals, drawers, menus) is Escape-closable, focus-trapped where modal, and returns focus to its opener.

### Intended visual deltas (don't flag as regressions)
- Dark mode: dark text on primary/destructive buttons (see flow 3).
- Bare `rounded` corners grew 0.25rem → 0.375rem (aligned to `--radius`).
- Minor padding/radius normalization where `BTN_*` constants now come from `buttonVariants`.
- Model-card entrance is shorter and smaller (300ms/8px → 200ms/4px) since it replays on search refinement.

## Verification
`typecheck`, `lint` (0 errors), `test` (162/162), `build` — all green after every wave. Per-plan feel-check scripts are in each `docs/plans/*.md`; executor deviations are recorded in the plan headers (002, 004).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013YpDb2toqctSFQV1AWQwqh


xiao-villamor:
Reviewed the motion work — the foundation is solid. Zero `transition-all`, zero `ease-in`, zero `scale(0)`, zero animated layout properties, zero ungated hover transforms across `frontend/src`. The overlay/menu/tab primitives are the right call.

I pushed four commits on top of your branch as [`0.9.0-design-motion`](https://github.com/xiao-villamor/PrintStash/compare/0.9.0...0.9.0-design-motion) rather than pushing into your draft. Cherry-pick or merge whatever you agree with — I'd like these in **your** PR rather than landing them behind you.

### 1. The Frontend CI failure — `fix(frontend): query model tabs by their ARIA tab role`

Making the tab strips a real ARIA `tablist` is correct, but it means the tabs are `role="tab"`, not `role="button"`. Both e2e suites still queried them as buttons and timed out on a locator that can no longer match. That's the *only* thing failing the Frontend job — two-line fix, suite goes green.

### 2. A real bug: the modal is torn out mid-exit — `fix(frontend): stop the modal being torn out mid-exit`

`ModalShell` passes `useMountTransition(open, 150)`, but its panel exits over `duration-fast` (**200ms**). The unmount timer fires 50ms early, so the panel and backdrop leave the DOM at ~75% through the scale+fade and closing a modal visibly snaps — the exact bug the primitives were introduced to fix.

`Drawer` (200/200) and `DropdownMenu` (150/150) happen to be correct, which is what makes this worth fixing at the seam rather than patching the number: `exitMs` and the `duration-*` class are two halves of one animation in two different languages, and nothing type-checks them against each other. I exported the scale as `DURATION` from `lib/overlay.ts` so all three overlays name the constant instead of hand-typing milliseconds.

### 3. The grid still enters as one flat block — `feat(frontend): stagger the model grid entrance`

Every card on a page runs `animate-card-in` simultaneously. A 30ms stagger, **capped at ten** (`PAGE_SIZE` is 60 — uncapped it would march for ~2s), so a full page still lands inside the 300ms budget. Done in CSS on the container rather than threading an index prop through `ModelCard`, which also picks up the collection cards sitting in the same grid.

One trap worth knowing: the stagger rules are `:nth-child` (specificity 0,2,0), so a reduced-motion override written as `.stagger-children > *` (0,1,0) **loses on specificity and silently does nothing**. I hit exactly that. Both halves are pinned by e2e tests now.

### 4. Five leftover raw durations — `refactor(frontend): put the last raw durations on the token scale`

`model-card` (200, 300), `printers-list` (200), `fab` (150), `bottom-nav-bar` (150) still hand-type durations next to the new scale. `grep -rE 'duration-[0-9]+' frontend/src/` is now empty.

### Deliberately not changed
- **Tabs indicator animates `width`** (`tabs.tsx:88`). Technically a layout property, but it's a 2px absolutely-positioned bar with no children so layout is contained, and `scaleX` would distort the `rounded-full` caps. Right tradeoff.
- **Toast easing.** Sonner only exposes colour vars; matching the house curve needs an `!important` fight with the dependency for something nobody consciously notices.

### Also on that branch
`DESIGN.md` — the token scale, the ten motion rules, the primitives, and a checklist for new UI, referenced from `AGENTS.md`. Your `docs/plans/` describe *how the migration was done*; this describes *what the system is*, so the drift you just removed doesn't come back. It documents your system, so please push back on anything I've mis-stated.

PR #62, opened 2026-08-04
v0.11.3: integrations, ingestion, and data-integrity hardening

### LINKED ISSUE #57: [Feature] Elegoo Centauri Carbon (SDCP): G-code upload and print-history import are both achievable — provider matrix rules out more than the firmware does
### Problem

Authored by AI (Claude, Opus 5 w. ultracode thinking), approved by me - a human :)

---

Running 0.11.1 against a Centauri Carbon (the original, not the CC2) on firmware V1.4.49, local SDCP v3 over WebSocket :3030. Live status has been solid — thanks for shipping SDCP at all.

`docs/provider-support.md` marks **Upload G-code**, **File Inventory** and **Print History Import** as unsupported for this printer, with the reason given as *"file-list probes can terminate"* the daemon and *"firmware exposes no safe file API"*.

That reason holds for the **file-list probe specifically**. It doesn't extend to upload or to history, and I have both working against the same firmware:

**Upload.** The failure people hit is real but it's packet size, not the operation. Large writes make the printer reset mid-transfer; chunked at **1 MB packets** it completes reliably — that's been our only upload path for ~50 prints across V1.4.46 and V1.4.49, with no daemon restarts. The CC1 accepts a plain `.gcode` file, no wrapper. Crucially, upload and the file-*list* command are independent: you can have the former without ever issuing the latter, which keeps the risky probe out of the picture.

**History.** SDCP Cmd 320 returns the printer's own history — 46 tasks on this machine, each with filename, begin/end times, status (`1` finished, `3` cancelled), layer count, and an md5 per file. No manual entry, which is the same low-friction property #47 credits Moonraker for. The md5 means an imported job can be matched to the exact stored revision rather than to a filename.

Three caveats that make this actively misleading if ignored:

- **No filament or spool data at all.** History says what ran and whether it finished; it cannot recover a spool attribution.
- **Some rows are dated 1970** — tasks predating the printer's RTC being set. We import those with null dates; importing them as 1970 silently corrupts any time-window statistic.
- **The printer returns history newest-first.** Taking the last element as "latest" inverts it. That mislabelled a file for us that was cancelled at layer 26 and then completed 620/620.

### Proposed solution

Smallest useful version: **enable chunked upload (1 MB packets) for the Elegoo Centauri provider, and leave File Inventory and Delete unsupported.**

That's a one-cell change to the matrix — `Upload G-code: ✗` → `✓ (chunked)` — and the safety argument is untouched, because it only ever applied to the file-list probe. It alone unblocks send-to-printer, which is the thing that currently forces SDCP users out of PrintStash at the last step.

History import via Cmd 320 is worth doing as a separate, later step: it's more code (parsing, the 1970 rows, the newest-first ordering) and it's useful without upload, so the two don't need to land together.

### Area

Printer integration

### Alternatives considered

Currently we drive the printer with our own SDCP client — it does the chunked upload and reads Cmd 320 — and then push the results into PrintStash through the REST API, creating models, G-code revisions and print jobs from the printer's history. That works and it's what we're doing today (46 print jobs and 38 revisions imported), but it means the vault is a passive mirror of a pipeline it can't participate in.

Also considered: uploading by hand through the printer's own UI (defeats the point of a vault), and posting this to Discussions as a hardware note — happy to move it there if you'd rather.

### LINKED ISSUE #58: [Bug] Spoolman integration — write_enabled defaults on where write-back is impossible, and the spool payload drops `location`
Authored by AI (Claude, Opus 5 w. ultracode thinking), approved by me - a human :)

---

### What happened?

Two independent problems found while connecting Spoolman to a non-Moonraker printer (Elegoo Centauri Carbon over SDCP). Filing together as they're both in the Spoolman integration; happy to split if you'd rather.

### 1. `write_enabled` defaults to `true` for providers that can never write back

`PUT /api/v1/spoolman` leaves `write_enabled` on unless it's explicitly set to false. Enabling the integration therefore turns on write-back for a provider where write-back can never happen — the docs are clear that non-Moonraker providers return no measured consumption, so PrintStash correctly refuses to decrement from a guess.

**Expected:** a flag that cannot do anything for the connected provider is off, or is gated on that provider supporting measured consumption.
**Actual:** it reads to the operator as "PrintStash is maintaining my inventory," when nothing is being written.

The risk isn't today, it's later. If write-back is ever extended to more providers, every existing install silently gains a second writer against a ledger that another tool may already own. Where something else is authoritative for filament, that surfaces as unexplained drift rather than an error. `write_force` is already correctly conservative — this is just the default on its neighbour.

### 2. `GET /api/v1/spoolman/spools` omits Spoolman's `location`

The returned payload is:

    [archived, color_hex, filament_id, filament_name, id, material,
     name, remaining_weight, used_weight, vendor_name]

Spoolman's own `location` field isn't passed through.

**Expected:** `location` available, so spools can be told apart.
**Actual:** for anyone with a multi-slot changer (AMS, CANVAS, MMU), the field that distinguishes otherwise identical spools is gone. I have three rolls of the same vendor, material and colour that differ only by which slot they're in; in the picker they are indistinguishable, and choosing the wrong one is silent.

`location` is free text in Spoolman, so passing it straight through would be enough.

### Steps to reproduce

**For 1 (write_enabled default):**
1. Have a Spoolman instance reachable from the container.
2. Configure a non-Moonraker printer (Elegoo Centauri Carbon over SDCP in my case).
3. Enable the Spoolman integration — set only `enabled` and `base_url`, either in Settings → Spoolman or via `PUT /api/v1/spoolman`.
4. Read the config back: `GET /api/v1/spoolman`.
5. `write_enabled` is `true`, though it was never set and cannot take effect for this provider.

**For 2 (missing location):**
1. In Spoolman, give two or more spools distinct `location` values (mine are `CANVAS-1`, `CANVAS-2`, `CANVAS-4`) while leaving vendor, material and colour identical.
2. Sync/enable the integration in PrintStash.
3. `GET /api/v1/spoolman/spools`, or open the spool picker in the UI.
4. The spools are indistinguishable — no `location` in the payload and nothing in the UI to separate them.

### Area

Other

### Version or commit

0.11.1

### Environment

Host OS: Windows 11 + WSL2 (Ubuntu 26.04)
Install method: Docker Compose, GHCR images, pinned 0.11.1, SQLite, local storage
Browser: Edge
Printer/provider: Elegoo Centauri Carbon (original, not CC2), firmware V1.4.49, local SDCP v3
Spoolman: 0.26.0, separate compose project, reached via host-gateway

### Relevant logs

```shell
# config readback after setting ONLY enabled + base_url
enabled       : True
base_url      : http://host.docker.internal:7912
write_enabled : True      <- never set by me
write_force   : False

# connectivity probe is healthy, so this is not a connection fault
POST /api/v1/spoolman/test
{"connected": true, "version": "0.26.0", "error": null, "native_hook_detected": false}

# spool payload - no `location`
GET /api/v1/spoolman/spools
{"id": 4, "filament_id": 2, "name": "PLA-CF", "filament_name": "PLA-CF",
 "vendor_name": "Elegoo", "material": "PLA", "color_hex": "1A1A1A",
 "remaining_weight": 1000.0, "used_weight": 0.0, "archived": false}
```

### LINKED ISSUE #61: [Bug]: Unable to import archive
### What happened?

I was trying to upload an archive and got the error

Image

### Steps to reproduce

1. Click on "Upload" button
2. Select "From zip" option
3. Select an archive

Please note, this is not happening for all archives just for specific cases.
The archive that throw the error in my case is about 63 MB in total, but inside there are nested folders for up to 5 levels, something similar to

`.zip root > Folder 1 > Subfolder 2 > Subfolder 3 > Subfolder 4 > Files`

I do not know exactly the total amount of files but I suspect it will be around 100 total files (both .stl as well as images)

### Area

File ingestion

### Version or commit

v0.11.2

### Environment

Host OS: Ubuntu 24.04
Install method: Docker Compose
Browser: Chrome

### Relevant logs

```shell

```

### LINKED ISSUE #64: [Feature]: Warn before sending a job the selected spool cannot cover
### Problem

Authored by AI (Claude - Opus 5 w. ultracode), approved by me - a human :)

---

PrintStash already knows how much filament a job needs and how much a spool has left, but nothing compares them before the file is sent.

`gcode_parser.py` extracts `filament_used_g` / `filament_weight` at ingest, and the Spoolman integration reads each spool's remaining weight. Today those two numbers only meet *after* the print, when `spoolman.py` decrements. So a job that cannot possibly finish is uploaded and started exactly like one that can.

I hit this on a multi-plate build: six plates, about 1446 g total, no single plate larger than a spool. Checking one job at a time waves four plates through and only fails partway into the fifth, roughly 22 hours in.

As far as I can tell no slicer or print manager does this — Spoolman itself is a passive ledger with no "will this fit?" concept, and its own plugins only decrement after the fact. It is a gap across the whole ecosystem, not something specific to my printer.

### Proposed solution

Smallest useful version: **on send, compare `filament_used_g` against the selected spool's remaining weight, and warn if it does not fit.** Nothing more — no blocking, no new data sources, no provider-specific code.

Three details that matter from having run this check for a while:

1. **Treat `remaining` of `null` as unknown, and let it fail closed.** Spoolman returns null when `initial_weight` is unset. A safety check that treats unknown as "plenty" is worse than no check. Same if Spoolman is unreachable.
2. **Allow checking several files as one set**, so a multi-plate build can be evaluated against one spool. This is the case that actually bit me; per-job checking cannot see it.
3. **Compare weight, not length.** Spoolman converts length server-side using the filament record's density and diameter, so a wrong density there silently skews the answer. The G-code already gives grams.

A later step, if this proves useful, would be to check that the file's `filament_type` and temperatures match the selected spool's material and `settings_extruder_temp` / `settings_bed_temp` — that catches sending a PETG file while a PLA spool is loaded. I am deliberately not asking for that here; it is a separate feature.

### Area

Printer integration

### Alternatives considered

I run this as a local pre-send gate today, so I am not blocked — but it lives outside PrintStash, which means it does not apply when I send from the UI, and that is exactly where I would want it.

Spoolman cannot do it: it has no concept of a pending job. A slicer cannot do it either, since it does not know which physical spool is loaded. PrintStash is the only place that holds both halves.

Happy to help implement it if you would like the contribution.

### LINKED ISSUE #65: Centauri Carbon: Mainboard ID is needed to start a print from an idle printer, but the field's helper text says it is only needed when paused or errored
Authored by AI (Claude - Opus 5 w. ultracode), approved by me - a human :)

---

**Area:** Other — there's no Elegoo/SDCP option in the dropdown. `Web UI` would fit the helper text on its own, but the defect is in the start path, so I've filed it free-form rather than picking a misleading area.

## What happened?

The **Mainboard ID** field on the Centauri Carbon setup form is marked *(recommended)* and its helper text reads *"Needed for reliable reconnection while paused or errored."*

That is narrower than the real requirement. The ID is also needed while the printer is **idle** — the state you start a print from. A user who reads that text, leaves the field blank and presses start gets a 5 s transport error instead of a print.

Tracing it:

- `printer_provider.py` passes `mainboard_id=printer.elegoo_centauri_mainboard_id` straight through, and only `host` is required, so the field can legitimately be empty.
- `ElegooCentauriClient.start()` goes through `_with_connection(True, … start_print(…))`.
- In `pycentauri` 0.9.0, `start_print()` calls `wait_for_mainboard()` (`client.py:425`), which waits up to 5 s for the printer to volunteer a frame carrying the MainboardID.

**Measured on the hardware today.** I tested the gate directly rather than guessing — the same `wait_for_mainboard()` call, reached through `attributes()` so that a test could never start a job. A print finished at 16:07:48; both samples are after it, machine idle:

| quiet for | passive listen, no command | `attributes()`, no mainboard_id | `attributes()`, mainboard_id set |
|---|---|---|---|
| 8.5 min | frame arrived at 2.06 s | **raised at 5.01 s** | succeeded, 0.13 s |
| 13.1 min | nothing in 5.01 s | **raised at 5.01 s** | succeeded, 0.16 s |

For contrast, the same script *during* the print: a frame arrived in 3.06 s and the call succeeded in 1.20 s. So the gate is satisfied while printing and not while idle.

**One correction to how I'd previously described this, including in `bjan/pycentauri#3`:** I'd said an idle Centauri Carbon volunteers no frames. That's too strong. At 8.5 min a frame *did* arrive for the passive listen, and then nothing arrived for the very next call seconds later. **The pushes are intermittent, not absent** — the machine only went reliably silent somewhere between 8.5 and 13 min after the job ended.

That makes the failure worse rather than better: it also fires *while* those residual frames are still coming, so it will look flaky rather than deterministic — easy to write off as the printer being temperamental.

Two things this does **not** affect, so they don't cost you time:

- **Upload.** `upload_file()` is plain chunked HTTP on port 80 and carries no MainboardID — matching what I measured in the earlier hardware run on #62.
- **`connect()`.** It opens the socket and returns without ever awaiting the ID. I'm flagging that explicitly because my earlier comment on #62 mentioned connecting to an idle printer in ~0 s. That measurement is real, but I'd withdraw the implication: `connect()` never waits in *any* state, so it tells you nothing about the idle case. The gate is per-command, not per-connection.

The helper text isn't wrong by accident — it mirrors pycentauri's own error string and `wait_for_mainboard()` docstring, which also say "paused or errored". Same wording gap, two projects.

## Steps to reproduce

```
1. Add a printer, provider Elegoo Centauri Carbon (elegoo_centauri_carbon).
2. Fill in the host. Leave "Mainboard ID" blank — the form allows it, and the
   helper text implies it only matters when paused or errored.
3. Leave the printer idle. It reproduces from roughly 8 minutes after a job
   ends, and reliably once the machine has gone quiet (~13 min in my case).
   It will NOT reproduce while a print is running.
4. Send a job with start_print: true.

Expected: the print starts.
Actual:   a 5 s transport error out of wait_for_mainboard(). The failure happens
          before anything goes on the wire, so no command reaches the printer and
          nothing is started — the failure is a clean no-op, not an ambiguous one.

Setting Mainboard ID makes it work every time.
```

## Version or commit

```
v0.11.2 (current release) — the helper text is identical on main and on branch
0.11.3, same file and lines, so this does not depend on #62 landing.
pycentauri 0.9.0.
```

## Environment

```
PrintStash: read from v0.11.2/main and branch 0.11.3 source. The hardware
            measurements below are at the pycentauri layer, NOT taken through a
            running PrintStash instance — see "scope" at the end.
Printer:    Elegoo Centauri Carbon 1 with CANVAS
Firmware:   V1.4.49
Host OS:    Windows 11, Python 3.13
```

## Relevant logs

```
--- idle, 13.1 min after the print ended -------------------------------
[pre] discovery
    machine_name    Centauri Carbon
    FIRMWARE        V1.4.49

[A] passive listen for 5.0s -- no command sent
    NOTHING after 5.01s -> PrinterError: printer did not push Attributes
    within 5.0s. Pass mainboard_id=... to Printer.connect() (discover()
    provides one) - the firmware does not push Attributes while paused or
    errored.

[B] attributes() with NO mainboard_id
    RAISED after 5.01s
    PrinterError: printer did not push Attributes within 5.0s. ...

[C] discover() then attributes() WITH mainboard_id
    SUCCEEDED in 0.16s -- pre-seeding works

--- idle, 8.5 min after the print ended --------------------------------
[A] FRAME ARRIVED after 2.06s        <- still settling: a frame did arrive
[B] RAISED after 5.01s               <- but the gate fired anyway
[C] SUCCEEDED in 0.13s

--- during the print (control) -----------------------------------------
[A] FRAME ARRIVED after 3.06s
[B] SUCCEEDED in 1.20s               <- gate never fires while printing
[C] SUCCEEDED in 0.16s
```

## Anything else

**Scope of what I'm claiming.** I measured the *gate* — `wait_for_mainboard()` — on the real printer, reached through `attributes()` rather than `start_print()` so that a test could never start a job. I have **not** watched PrintStash's own `start()` fail: my 0.11.3 run used `start_print: false` by design. The path from `start()` → `start_print()` → `wait_for_mainboard()` is right there in the source, so I think the inference is tight, but I'd rather state it as an inference than overclaim.

The cheap fix is the helper text — *"while idle, paused, or errored"* — which is also fix (1) I suggested on the pycentauri issue. A larger option, if you'd rather `start` just worked with the field blank, is to fall back to UDP discovery: `pycentauri.discover()` returns the MainboardID (it also returns the firmware version, which is how I got V1.4.49 above), and there's currently no `discover()` call anywhere in the repo.

Happy to test any patch — I have the hardware in this state routinely.

xiao-villamor:
### Release fixes and integration work

- Spoolman write-back now defaults to off for new configurations, and spool locations are exposed by the API and picker so otherwise-identical spools can be distinguished.
- Zip import counts actual files rather than directory records against the archive-entry limit.
- Elegoo Centauri Carbon upload is enabled through pycentauri 0.9.1, with the CC1 and CC2 hardware-validation limits documented.
- The send flow warns when the selected Spoolman spool has insufficient or unknown remaining material.
- Centauri Carbon setup and settings now explain that Mainboard ID is needed for reliable printer commands while idle, paused, or errored.

### Data-integrity and reliability audit follow-up

- Artifact revision numbers are now reserved atomically. A new migration repairs pre-existing duplicate versions/recommendation markers, backfills the per-model counter, and installs database uniqueness guards.
- SQLite backups now use a transactionally consistent online snapshot, include committed WAL data, derive blob ownership from that same snapshot, validate integrity, and stream the database into the archive.
- Restore now blocks new mutations, drains admitted writes, stages and validates the database/blobs, rolls storage changes back on failure, and restores SQLite through its online backup API instead of replacing the live file.
- Background writers participate in the restore maintenance gate so GC, external scans, notifications, printer status updates, and fleet dispatch cannot race a restore.
- Unversioned databases are stamped at Alembic head only when their schema demonstrably matches the current metadata; incomplete orphan schemas now fail closed.
- OctoPrint multipart uploads stream the source file instead of reading a potentially large G-code entirely into memory.
- Recommendation changes flush the old marker before promoting a replacement, matching the new partial unique index.

Fixes #57. Fixes #58. Fixes #61. Fixes #64. Fixes #65.

File inventory, deletion, and print-history import for Elegoo Centauri stay unsupported. History import still requires restructuring job_import.py away from its Moonraker-only assumptions.

## Testing

The CI-equivalent matrix was run locally before the metadata-only pycentauri 0.9.1 bump. On the final branch state, the strict dependency audit, lock verification, Ruff, and the Centauri/provider suite (250 passed, 1 skipped) were rerun locally. The earlier GitHub CI run passed every job except the dependency audit, which now passes locally:

- [x] Backend full suite: 2,125 passed, 22 skipped; 97% coverage (95% required)
- [x] Backend E2E included in the full suite
- [x] Ruff and Bandit
- [x] Real migration upgrades from v0.7.0 and v0.11.2 to head; the upgraded app reported v0.11.3 and returned HTTP 200
- [x] S3/MinIO suite: 85 passed
- [x] Frontend unit suite: 234 passed
- [x] Release metadata and in-app changelog consistency: 5 passed
- [x] Frontend typecheck, lint, and production build
- [x] Mock-API Playwright: 17 passed
- [x] Real-backend Playwright: 58/59 on the first run; the single printer-preset timeout passed on isolated rerun
- [x] Native Docker images built; the container health endpoint passed a smoke test
- [x] pip-audit --strict: no known vulnerabilities found
- [ ] ARM64/QEMU image validation was not run locally because the runner lacked enough disk space

Two backend printer-emulator timing failures also passed immediately on isolated rerun.

## Notes

- Version 0.11.3 is an explicit patch-policy exception because it includes two user-facing additions and an append-only migration; the release notes tell operators to back up before upgrading.

- The Spoolman write-default change affects only newly created SystemConfig rows; existing stored settings are unchanged.
- SpoolRead.location and the Elegoo can_upload capability are additive API changes.
- Elegoo Centauri upload remains beta. CC1 upload has been exercised end to end on real hardware by the reporter, but the supervised active-print checklist is incomplete; CC2 upload remains unvalidated on hardware.
- The Mainboard ID remains optional because HTTP upload does not need it. The UI now accurately identifies it as necessary for reliable command paths in quiet printer states.


InfiniteBSOD:
Authored by AI (Claude - Opus 5 w. ultracode), approved by me - a human :)

---

Hardware validation from an actual Centauri Carbon 1, since the upload path currently ships as beta against mocked transports.

**What I ran.** I uploaded a real 7,188,340-byte sliced G-code to a physical CC1 through `pycentauri` 0.9.0's `upload_cc1()` — the call `ElegooCentauriService.upload()` delegates to:

- 7,188,340 bytes in **3.3 s** (~2,126 KiB/s)
- **7 chunks** at the 1 MiB `CHUNK_SIZE`, 7 progress callbacks
- returned the remote filename
- SDCP Cmd 258 afterwards lists the file at `/local/<name>.gcode`

I also exercised the connection your `upload()` opens — `Printer.connect(host, enable_control=True)` — against an **idle** printer (machine status 0), which is the awkward case for this machine. It connected in ~0 s.

**Please scope this precisely when you decide about beta.** I did **not** run PrintStash's own code: #62 is unmerged, no release carries it, and there's no `0.11.3` image published, so building the branch and migrating my library DB wasn't a trade I wanted to make on a working instance. So this validates *the dependency call and the connection*, which is what mocks can't reach — not your service layer, Vault retrieval, or the capability gating. If that's enough to move the transport out of beta, great; if you'd rather I ran the branch end-to-end, say so and I'll set up an isolated instance.

Two things I checked and want to explicitly clear, so they don't cost you time:

- `Printer.upload_file()` raises `ControlDisabledError` unless `enable_control=True`. Your `upload()` passes `True`, so this is fine — flagging only because the test double has no such gate, so it wouldn't surface in CI.
- `pycentauri` posts to port **80**; my own long-standing uploader has always used **3030**. Both ports are open on the CC1 and port 80 worked, so this is a difference rather than a defect.

---

**Separately — the history import from #57.**

Now that upload is landing, the remaining gap for SDCP printers is print-history import, which didn't ship because `job_import.py` is built around Moonraker rather than the provider abstraction. `pycentauri` 0.9.0 already exposes `print_history()` (Cmd 320 + 321), so the data is reachable through the same dependency you just bumped.

I have 46 real tasks off this printer and I'm happy to hand over whatever shape is useful. The three gotchas that would otherwise bite an implementation, up front:

1. **No filament data.** SDCP history returns no material or consumption figures at all, so imported jobs can't populate filament usage or cost — they'd need to be nullable.
2. **Pre-RTC rows are dated 1970.** Tasks recorded before the printer's clock syncs carry epoch-zero timestamps. They must import as null rather than as 1970 dates, or they'll wreck any date-range statistics.
3. **The printer returns history newest-first**, which is the opposite of what an incremental importer usually assumes.

Happy to supply a redacted dump of the 46 tasks (filename, begin/end, status, layer count, md5) if that would help — just say what format suits.

xiao-villamor:
Thanks, this is exactly the kind of check mocks can't give us.

Fixed one gap this exposed: our test double (FakeConnection) only enforced enable_control through a manual assert in each test's connector, not in the fake itself, so a regression that dropped enable_control=True from a write path like upload() wouldn't reliably fail in CI. It now mirrors pycentauri's real _require_control gate on every control method, plus a dedicated regression test for it (commit 64992c5).

On scoping, I agree with your read. You validated the pycentauri dependency call and the connection, which is real and useful, but this repo's own rule says mocked-transport tests don't count as hardware validation, and a Hardware Validation Log entry needs the fuller smoke test in manual-testing.md, not just the dependency call. So I'm holding off on a log row for now. I wrote up what you found in docs/provider-support.md and known-limitations.md instead, so it isn't silently implied as fully validated.

While updating those docs I found something this indirectly surfaced: capabilities are advertised at the provider level, shared by both Centauri Carbon variants, so enabling upload turned it on for CC2 too, even though only CC1 has any real-world confirmation. CC2's upload path (upload_cc2, PUT with Content-Range) is a different code path in pycentauri, and nobody has run it against real CC2 hardware. manual-testing.md's CC2 checklist item also still said to confirm upload stays unavailable, which is now stale. Fixed both (commit 8c1dfdb).

Port 80 vs 3030: noted, no action needed since you called it a difference, not a defect.

If you're up for setting up that isolated instance to run 0.11.3 end to end, that would be enough to move upload out of beta for CC1 at least, and CC2 too if you have one to test against. No pressure either way.

On history import: yes to the redacted dump, whatever format is easiest, CSV or JSON or whatever you already have. Filename, begin/end, status, layer count, md5 is exactly what's useful. I'll factor in the three gotchas (no filament data, null instead of 1970 dates, newest-first ordering) when it gets built. job_import.py needs restructuring off its Moonraker-only assumptions first regardless.

InfiniteBSOD:
Authored by AI (Claude - Opus 5 w. ultracode), approved by me - a human :)

---

I ran 0.11.3 end to end against the real printer, in an isolated instance so my live library was never touched.

**Setup**

- branch 0.11.3 built from source, separate compose project, own volumes, empty database, port 7914
- printer added as `elegoo_centauri` / `elegoo_centauri_carbon`
- my normal 0.11.1 instance kept running untouched the whole time

**Result: upload works.**

- `POST /api/v1/printers/1/send` with `start_print: false` returned **HTTP 200**, PrintJob `state: completed`
- file: 7,188,340 bytes, sent as `printstash_upload_test.gcode`
- I verified it independently rather than trusting the response: an SDCP file listing shows `/local/printstash_upload_test.gcode` on the printer, and the machine stayed idle with print status 0, so nothing was started
- `detected_model` came back as "Elegoo Centauri Carbon", so the provider identified the machine itself
- capabilities reported `can_upload: true`, `can_list_files: false`, `support_level: beta`

This is your own service layer this time — vault retrieval, capability gating, the ingest → send path — not just the pycentauri call.

**What I did not do, so please scope the log row accordingly.**

Your CC1 row in `manual-testing.md` lists six things: SDCP status, temperatures, chunked upload, supervised start-existing-file, pause/resume/cancel, and reconnect-while-paused using the saved mainboard ID.

I covered **status and chunked upload**. I did not run the four that need an active print.

That is a hard blocker rather than a choice: this printer currently cannot start a print at all. It fails at layer 0 reporting a filament runout on a new spool, which I have open with Elegoo support. So the supervised items are not runnable here until that is fixed. I am happy to run them once it is — I would like to drive prints through PrintStash going forward, so testing that path properly is in my interest too.

If upload alone is enough for a partial entry, great. If the row needs all six, it will have to wait, and I would rather tell you that than have it recorded as fully validated.

**CC2**: I do not have one, so I cannot help with the `upload_cc2` path.

**History dump**: ready whenever you want it — 49 tasks as JSON with filename, begin/end, status, layer counts and md5, in the printer's native newest-first order. One extra gotcha I found while preparing it: `error_status_reason` is `0` even on a job that failed to start, so it cannot be used to detect failure. Status alone carries that.

<details>
<summary><b>Redacted CC1 print history — 49 tasks (JSON)</b></summary>

Native newest-first order, not re-sorted, so the ordering gotcha is visible in the data. `TaskId` and `TimeLapseVideoUrl` removed — not needed for import.

```json
{
 "_source": "Elegoo Centauri Carbon (CC1), firmware V1.4.49, SDCP Cmd 320 + 321 via pycentauri-equivalent client",
 "_exported": "2026-08-05",
 "_order": "AS RETURNED BY THE PRINTER = NEWEST FIRST. Not re-sorted, so the ordering gotcha is visible.",
 "_gotchas": [
  "No filament data of any kind. SDCP history returns no material, no length, no weight. Imported jobs cannot populate filament usage or cost - those fields must be nullable.",
  "Pre-RTC rows carry epoch-zero timestamps (4 of 49 here). They must import as null, not as 1970 dates, or date-range statistics break.",
  "Printer returns newest-first, the opposite of what an incremental importer usually assumes.",
  "TaskStatus: 1 = completed, 3 = cancelled/stopped.",
  "error_status_reason is 0 even on a job that failed to start, so it cannot be used to detect failure."
 ],
 "_redacted": "TaskId (device UUID) and TimeLapseVideoUrl removed - not requested and not needed for import.",
 "task_count": 49,
 "tasks": [
  {
   "filename": "/local/box_dividers_placf.gcode",
   "begin_time": 1785933318,
   "end_time": 1785933585,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 0,
   "layers_total": 410,
   "md5": "775f1927e7bb62f05c28345710934b4d"
  },
  {
   "filename": "/local/box_dividers_placf.gcode",
   "begin_time": 1785932293,
   "end_time": 1785933013,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 0,
   "layers_total": 410,
   "md5": "775f1927e7bb62f05c28345710934b4d"
  },
  {
   "filename": "/local/box_dividers_placf.gcode",
   "begin_time": 1785931098,
   "end_time": 1785931752,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 0,
   "layers_total": 410,
   "md5": "775f1927e7bb62f05c28345710934b4d"
  },
  {
   "filename": "/local/led_bracket_brim_support_slot2.gcode",
   "begin_time": 1785764402,
   "end_time": 1785765197,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 35,
   "layers_total": 35,
   "md5": "696c7c3f681b9325207858e55afdf397"
  },
  {
   "filename": "/local/clip_support_ingroove_ba90_slot2.gcode",
   "begin_time": 1785694194,
   "end_time": 1785695881,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 95,
   "layers_total": 95,
   "md5": "a1f86c34124f78953035cbd1af3878a3"
  },
  {
   "filename": "/local/reloc_dock_standing_slot4.gcode",
   "begin_time": 1785686204,
   "end_time": 1785693342,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 520,
   "layers_total": 520,
   "md5": "335a57b71f562e50bf702a910495f4de"
  },
  {
   "filename": "/local/s4feet_plus_clip.gcode",
   "begin_time": 1785664363,
   "end_time": 1785672989,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 85,
   "layers_total": 85,
   "md5": "505a311e5e4b5297b3fec9aa22ecde97"
  },
  {
   "filename": "/local/s4_stand_75_nobrim.gcode",
   "begin_time": 1785588756,
   "end_time": 1785613991,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 380,
   "layers_total": 380,
   "md5": "1bc32b2cbc84cb74b08a298f455c6d1c"
  },
  {
   "filename": "/local/s4_stand_75.gcode",
   "begin_time": 1785585032,
   "end_time": 1785587856,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 6,
   "layers_total": 380,
   "md5": "b564fd8e949387ee51777600ffc0db87"
  },
  {
   "filename": "/local/reloc_lower_plus_dock.gcode",
   "begin_time": 1785574823,
   "end_time": 1785582301,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 79,
   "layers_total": 79,
   "md5": "f3254f7d4c7ac081bcc4b2dad6b71be7"
  },
  {
   "filename": "/local/start_test_coupon.gcode",
   "begin_time": 1785573177,
   "end_time": 1785573616,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 30,
   "layers_total": 30,
   "md5": "dbb4aa98876f43d6d330c9139aa2c387"
  },
  {
   "filename": "/local/noled_slim2_rr_combo.gcode",
   "begin_time": 1785499855,
   "end_time": 1785522998,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 525,
   "layers_total": 525,
   "md5": "36b796a95956d33dca3f848932353e30"
  },
  {
   "filename": "/local/noled_p3_thick.gcode",
   "begin_time": 1785482111,
   "end_time": 1785498872,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 525,
   "layers_total": 525,
   "md5": "0d57fca0a44e5403b30011534cd7083c"
  },
  {
   "filename": "/local/noled_p3_thick.gcode",
   "begin_time": 1785433342,
   "end_time": 1785436209,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 64,
   "layers_total": 525,
   "md5": "27c5e645410a23c6dd9c7f3753cc8189"
  },
  {
   "filename": "/local/noled_p1_corners3.gcode",
   "begin_time": 1785411170,
   "end_time": 1785430202,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 525,
   "layers_total": 525,
   "md5": "ff53e4d976fbf964815781e3c2cc735b"
  },
  {
   "filename": "/local/vent_riser_v2.gcode",
   "begin_time": 1785326278,
   "end_time": 1785383410,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 222,
   "layers_total": 222,
   "md5": "04865f4e4c7946aaedfe02bf7b451ccd"
  },
  {
   "filename": "/local/reloc_p3_corners_v2.gcode",
   "begin_time": 1785259279,
   "end_time": 1785273287,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 525,
   "layers_total": 525,
   "md5": "cf0c0aea7e46476c2ab1bc7c36348452"
  },
  {
   "filename": "/local/reloc_right_upper_vert.gcode",
   "begin_time": 1785246498,
   "end_time": 1785256778,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 1018,
   "layers_total": 1018,
   "md5": "3d577d8caf1373bff8137fa342abe8d7"
  },
  {
   "filename": "/local/ptfe_tool_standing.gcode",
   "begin_time": 1785243412,
   "end_time": 1785245613,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 398,
   "layers_total": 398,
   "md5": "3eb2b53143fe547f20f58536c709c19e"
  },
  {
   "filename": "/local/led_bracket_only.gcode",
   "begin_time": 1785242360,
   "end_time": 1785243123,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 25,
   "layers_total": 25,
   "md5": "375fec334ec6c01e71bf970e54d62183"
  },
  {
   "filename": "/local/combo4_bracket_tool.gcode",
   "begin_time": 1785240778,
   "end_time": 1785242053,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 53,
   "layers_total": 100,
   "md5": "196cc589d9623a26c93a209c188d0151"
  },
  {
   "filename": "/local/combo3_bracket_coupon_ptfetool.gcode",
   "begin_time": 1785234563,
   "end_time": 1785237082,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 87,
   "layers_total": 100,
   "md5": "108f234e36735b4d70da6acefd324671"
  },
  {
   "filename": "/local/reloc_p2_slim.gcode",
   "begin_time": 1785163679,
   "end_time": 1785182213,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 525,
   "layers_total": 525,
   "md5": "4674260cc56b1487298ac5f33fc2856b"
  },
  {
   "filename": "/local/reloc_p3_corners.gcode",
   "begin_time": 1785144425,
   "end_time": 1785161121,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 799,
   "layers_total": 799,
   "md5": "0d9d7ead7684b281a323cd1dd6e22b39"
  },
  {
   "filename": "/local/reloc_p3_corners.gcode",
   "begin_time": 1785144005,
   "end_time": 1785144265,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 1,
   "layers_total": 799,
   "md5": "0d9d7ead7684b281a323cd1dd6e22b39"
  },
  {
   "filename": "/local/reloc_corner_fl.gcode",
   "begin_time": 1785075449,
   "end_time": 1785082413,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 525,
   "layers_total": 525,
   "md5": "703d1b843187b42a63d837c8ddba3752"
  },
  {
   "filename": "/local/3DBenchy.petg-cf.gcode",
   "begin_time": 1785075383,
   "end_time": 1785075441,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 0,
   "layers_total": 240,
   "md5": "8221c8f2504d55d67a515115240f744c"
  },
  {
   "filename": "/local/reloc_corner_fl.gcode",
   "begin_time": 1785075019,
   "end_time": 1785075180,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 1,
   "layers_total": 525,
   "md5": "703d1b843187b42a63d837c8ddba3752"
  },
  {
   "filename": "/local/reloc_dock_trio.gcode",
   "begin_time": 1784811208,
   "end_time": 1784821730,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 552,
   "layers_total": 552,
   "md5": "ea245d30751106097599b8067042eadd"
  },
  {
   "filename": "/local/01_front_top_solo.gcode",
   "begin_time": 1784732462,
   "end_time": 1784742068,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 620,
   "layers_total": 620,
   "md5": "2457912eaa1ce8ad4b247662dcfc9ba0"
  },
  {
   "filename": "/local/01_front_top_solo.gcode",
   "begin_time": 1784731422,
   "end_time": 1784732132,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 26,
   "layers_total": 620,
   "md5": "2457912eaa1ce8ad4b247662dcfc9ba0"
  },
  {
   "filename": "/local/02_front_down.gcode",
   "begin_time": 1784721472,
   "end_time": 1784731140,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 750,
   "layers_total": 750,
   "md5": "fa713dab16cf376eed4bd9448bc72c36"
  },
  {
   "filename": "/local/stringrig_B_pa052.gcode",
   "begin_time": 1784718835,
   "end_time": 1784720672,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 200,
   "layers_total": 200,
   "md5": "053c08ab70c1cead63817e9b56aaf1ab"
  },
  {
   "filename": "/local/stringrig_A_control.gcode",
   "begin_time": 1784716798,
   "end_time": 1784718676,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 200,
   "layers_total": 200,
   "md5": "dd58f27e9978723a76bb05baace317d2"
  },
  {
   "filename": "/local/01_front_top_FIXED.gcode",
   "begin_time": 1784633575,
   "end_time": 1784643190,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 620,
   "layers_total": 620,
   "md5": "2dbc61fb622f080af378be3e60b414fb"
  },
  {
   "filename": "/local/01_front_top_FLIPPED.gcode",
   "begin_time": 1784629804,
   "end_time": 1784630454,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 22,
   "layers_total": 620,
   "md5": "88457efd07f09d9c9834a7854f3f7cc6"
  },
  {
   "filename": "/local/01_front_top.gcode",
   "begin_time": 1784573105,
   "end_time": 1784581219,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 567,
   "layers_total": 620,
   "md5": "896733a13fd2b6d626c02b71bc4492f4"
  },
  {
   "filename": "/local/riser_plate2_v4_brim.petg-cf.gcode",
   "begin_time": 1784571337,
   "end_time": 1784572772,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 22,
   "layers_total": 750,
   "md5": "77222461a6e209f58f84bee3f3a59eee"
  },
  {
   "filename": "/local/riser_plate2_v3.petg-cf.gcode",
   "begin_time": 1784568602,
   "end_time": 1784571174,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 30,
   "layers_total": 750,
   "md5": "b836a75587c9ec2b35ec17ef0258c3a6"
  },
  {
   "filename": "/local/stringtest_240.petg-cf.gcode",
   "begin_time": 1784566569,
   "end_time": 1784568115,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 203,
   "layers_total": 203,
   "md5": "1eb9617367319b25b14b8c83738563b6"
  },
  {
   "filename": "/local/stringtest_250.petg-cf.gcode",
   "begin_time": 1784564701,
   "end_time": 1784566258,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 203,
   "layers_total": 203,
   "md5": "b7d1792c9e4e45c569da053515c43c7b"
  },
  {
   "filename": "/local/riser_plate2_v2.petg-cf.gcode",
   "begin_time": 1784563208,
   "end_time": 1784563961,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 18,
   "layers_total": 750,
   "md5": "7892f098a6f3490983842c262593ead5"
  },
  {
   "filename": "/local/riser_plate2_front.petg-cf.gcode",
   "begin_time": 1784561975,
   "end_time": 1784563047,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 16,
   "layers_total": 750,
   "md5": "d318b440f82f5303537f7ad7232d5793"
  },
  {
   "filename": "/local/riser_plate2_front.petg-cf.gcode",
   "begin_time": 1784560963,
   "end_time": 1784561791,
   "status": 3,
   "error_status_reason": 0,
   "layers_printed": 17,
   "layers_total": 750,
   "md5": "d318b440f82f5303537f7ad7232d5793"
  },
  {
   "filename": "/local/s4_adapter_set_v2.petg-cf.gcode",
   "begin_time": 1784553343,
   "end_time": 1784560476,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 365,
   "layers_total": 365,
   "md5": "f868d4d24f4a7eb0f97d7ae78c7bb5c5"
  },
  {
   "filename": "/local/s4_adapter_set.petg-cf.gcode",
   "begin_time": 183748,
   "end_time": 190872,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 365,
   "layers_total": 365,
   "md5": "30e98b641aa29de1e73a80a7f500e2d3"
  },
  {
   "filename": "/local/s4_adapter_set.petg-cf.gcode",
   "begin_time": 176133,
   "end_time": 183265,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 365,
   "layers_total": 365,
   "md5": "30e98b641aa29de1e73a80a7f500e2d3"
  },
  {
   "filename": "/local/s4_adapter_set.petg-cf.gcode",
   "begin_time": 10118,
   "end_time": 17244,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 365,
   "layers_total": 365,
   "md5": "30e98b641aa29de1e73a80a7f500e2d3"
  },
  {
   "filename": "/local/3DBenchy.petg-cf.gcode",
   "begin_time": 7086,
   "end_time": 9418,
   "status": 1,
   "error_status_reason": 0,
   "layers_printed": 240,
   "layers_total": 240,
   "md5": "8221c8f2504d55d67a515115240f744c"
  }
 ]
}
```

</details>

PR #78, opened 2026-08-20
docs: 0.12.0 is published, not pending

jorgehermo9:
`UPGRADE.md` still opens its 0.12.0 section with:

> The `0.12.0` branch is not a published release yet. Its upgrade-relevant changes are documented
> here so the pull request can be reviewed without implying that an image or tag already exists.

The tag published on 2026-08-19, so that framing now tells an operator the opposite of what is
true, on the page they read while deciding whether to upgrade.

The bullets are unchanged. Only the heading and the preamble move from pending to published.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01T2rwPqvcAYVX9pmqfvw8sr

PR #79, opened 2026-08-20
feat: complete 0.12.2 imports, previews, and Bambu reproducibility

### LINKED ISSUE #67: [Bug]: Thumnails not generated
### What happened?

After I uploaded a model the thumbnail of such model is not generated as seen in the screenshot

Image

If I open a single model, then the 3d file is properly displayed

Image

I'd expect the thumbnail to be created right after the model is uploaded, or I've misunderstood that part?

### Steps to reproduce

1. Create a new collection
2. Upload an STL file
3. Review the collection contents

PS: in this specific case, I've used the "Bulk" upload option, not sure if this is related somehow

### Area

Other

### Version or commit

v0.11.3

### Environment

Host OS: Ubuntu 26.04
Install method: Docker Compose (based on the light version)
Browser: Google Chrome

### Relevant logs

```shell

```

### LINKED ISSUE #69: [Bug]: Bambu LAN on Bambu A1 fails with Paho v2 callbacks, PUBACK handling, cleanup order, and partial MQTT reports
### What happened?

WHAT HAPPENED?

I tested PrintStash 0.11.4 extensively with a physical Bambu Lab A1 using the Bambu LAN (beta) provider.

The printer was reachable and direct tests confirmed that TCP, TLS, MQTT authentication, Bambu certificate/serial validation, SUBSCRIBE, `pushall`, and incoming print reports all worked correctly. However, PrintStash still reported the printer as OFFLINE/UNKNOWN.

During debugging I found four separate issues in the Bambu provider:

1. Paho MQTT v2 `ReasonCode` incompatibility

The callback uses:

```python
if int(reason_code) != 0:
```

With Paho MQTT callback API v2, `reason_code` is a `ReasonCode` object and this raises:

```text
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'ReasonCode'
```

Changing it to:

```python
if reason_code != 0:
```

resolves this.

2. MQTT cleanup order can hang

The current cleanup order is:

```python
client.loop_stop()
client.disconnect()
```

In my tests `loop_stop()` could block waiting for the network thread while the MQTT connection remained open.

Reversing the order fixed it:

```python
client.disconnect()
client.loop_stop()
```

3. Bambu A1 processes the QoS 1 PUBLISH but no PUBACK is observed

PrintStash currently waits for:

```python
info.wait_for_publish(timeout=timeout)
if not info.is_published():
    raise ProviderError("bambu_command_not_published", ...)
```

In a direct MQTT test, the A1 received and processed the command and returned multiple valid `print` responses, but Paho never observed a PUBACK.

Test result:

```text
CONNACK: Success
SUBACK received
PUBLISH sent
PRINT RESPONSE received
response_received: True
puback_received: False
```

Therefore PrintStash raised `bambu_command_not_published` even though the command had actually succeeded.

I changed the check to verify only that Paho accepted the outgoing publish:

```python
if info.rc != mqtt.MQTT_ERR_SUCCESS:
    raise ProviderError(...)
```

and then waited for the actual Bambu response.

4. Partial Bambu MQTT reports overwrite a valid printer state with UNKNOWN

Bambu sends partial `print` reports which do not always contain `gcode_state`.

The current normalization code creates a `print_stats.state` value even when `gcode_state` is absent, effectively producing an empty state. The printer hub then maps this to UNKNOWN and overwrites the previous READY state.

I changed `_normalize_status()` so that fields are only added when they are actually present in the incoming report, for example:

```python
if "gcode_state" in print_report:
    ...
```

The same approach is used for `mc_percent`, `subtask_name`, `print_error`, etc.

After applying these four changes, the Bambu A1 integration became stable.

I then built a derived image from:

```text
ghcr.io/xiao-villamor/printstash-api:0.11.4
```

containing only the patched `printer_provider.py`, recreated the API container with `--force-recreate`, and verified that the fix survives container recreation.

Final state after recreation:

```text
printer: Bambu A1
status: READY
last_error: None
```

`last_seen_at` continues to update normally.

A direct `pushall` test also returned:

```text
command: push_status
gcode_state: FINISH
mc_percent: 100
```

So the full A1 MQTT path is working after these changes.

No printer access code, serial number, private URL, or other credentials are included here.

I would be happy to provide the exact diff or test individual fixes separately if useful.

### Steps to reproduce

1. Run PrintStash 0.11.4 with `paho-mqtt >= 2.1`.
2. Add a physical Bambu Lab A1 using the Bambu LAN (beta) provider.
3. Configure a valid LAN access code and printer serial.
4. Confirm that TCP port 8883 is reachable.
5. Start/restart the PrintStash API and let the printer worker connect.
6. Observe the backend logs and printer status.

Observed sequence before the fixes included:

- `TypeError` from `int(reason_code)`
- after fixing that, MQTT request cleanup could hang
- after fixing cleanup, `bambu_command_not_published` occurred even though the A1 returned valid print responses
- after fixing publish handling, the printer initially became READY but later changed to UNKNOWN when partial MQTT reports without `gcode_state` arrived

Applying the four changes described above resulted in a stable READY state with no `last_error`.

### Area

Bambu LAN

### Version or commit

0.11.4

### Environment

Host OS: Ubuntu 26.04 LTS
Kernel: 7.0.0-29-generic
Architecture: x86_64

Deployment: Docker / Docker Compose
Docker Compose: 2.40.3+ds1-0ubuntu1

PrintStash: 0.11.4
API image: ghcr.io/xiao-villamor/printstash-api:0.11.4

Printer: Bambu Lab A1
Provider: Bambu LAN (beta)
Connection: local LAN MQTT/TLS on port 8883

Spoolman is connected, but is not involved in reproducing this issue.

### Relevant logs

```shell

```

### LINKED ISSUE #70: [Feature]: Preserve the exact printed artifact/revision and slicer settings for Bambu external print jobs
### Problem

My main reason for using PrintStash is reproducibility, not only a list of past prints.

I often make a small but important change in Bambu Studio — for example scale, orientation, supports, wall count, infill, layer height, or another slicer setting — print the part successfully, and then forget to save that exact project state.

Weeks later I may want to print the same part again. At that point, the most useful history entry would be the exact artifact/revision that produced the successful physical print, together with the slicer settings used.

With Bambu LAN, PrintStash can detect an externally initiated print and create an external PrintJob. However, the current Bambu status normalization uses:

```python
subtask_name or project_id
```

as the filename-like value.

On my physical Bambu Lab A1, a completed job returned for example:

```text
command: push_status
gcode_state: FINISH
mc_percent: 100
subtask_name: 0.2mm layer, 2 walls, 100% infill
```

That `subtask_name` is useful metadata, but it is not the original 3MF/G-code filename and it is not enough to recover the exact printed revision.

So today an external Bambu Studio print can appear in history while still losing the information that matters most for reproducing a known-good print.

This is closely related to the goals described in issue #47 ("Print History and Job Tracking"), especially G-code revision tracking and slicer metadata.

A separate Bambu LAN bug report from the same physical A1 testing is #69.

### Proposed solution

For external Bambu jobs, preserve as much of the exact printed state as the Bambu LAN protocol makes available.

Smallest useful version:

1. Keep creating an external PrintJob as today.
2. Preserve the best available real file/task identity instead of treating a generic `subtask_name` as the original filename.
3. If the exact printed 3MF/G-code artifact can be retrieved from the printer, store an immutable copy/revision and link the PrintJob to it.
4. If the exact artifact cannot be retrieved, store the best available metadata and clearly mark the job as metadata-only.

Where available, useful metadata would include:

- original filename / project name;
- Bambu task/project identifiers;
- subtask name;
- slicer and slicer version;
- layer height;
- wall count;
- infill;
- object scale;
- plate/orientation information;
- support settings;
- filament/material profile;
- nozzle/profile information;
- printer profile;
- timestamps;
- any stable remote file/task identifier that can later correlate the job with a file.

The UI should clearly distinguish between:

- fully reproducible job — exact immutable artifact/revision stored;
- partially reproducible job — metadata captured, artifact unavailable;
- basic external job — only limited status/history information available.

A user should never mistake `subtask_name` for the archived printable file.

The desired outcome is a "known good print memory": I should be able to open a successful historical job and answer:

- What exactly did I print?
- Which revision was it?
- What scale/settings did it use?
- Can I reproduce the same result later?

### Area

Printer integration

### Alternatives considered

The current workaround is to remember to manually save every Bambu Studio project or export the final 3MF/G-code before printing.

That works only if the user remembers to do it, which defeats much of the value of automatic print history.

Another workaround is to keep only the external PrintJob metadata currently exposed by MQTT. This is useful for basic history, but it does not reliably preserve the exact known-good artifact or settings that produced the physical part.

If retrieving the exact artifact from the Bambu printer is technically impossible, a metadata-only fallback would still be valuable as long as PrintStash clearly shows that the artifact itself was not archived.

TESTED WITH

PrintStash: 0.11.4
Printer: Bambu Lab A1
Provider: Bambu LAN (beta)
Deployment: Docker / Docker Compose

No printer access code, serial number, private URL, or other credentials are included.

### LINKED ISSUE #82: PRD: Bambu LAN reproducibility — exact artifact capture + duplicate reconciliation (#70)
## 1. Problem / the ask

PrintStash 0.12.0 shipped the Bambu LAN (beta) external-print recorder: externally started prints (e.g. from Bambu Studio) become PrintJob rows carrying the printer-reported identity fields, and a best-effort FTPS attempt to archive the exact G-code/project from the printer's cache. A real-hardware follow-up on a physical Bambu Lab A1 (issue #70, reporter lbudzynowski, 2026-08-22) confirmed two defects in the shipped feature:

1. **Exact artifact capture fails** on a real print: `artifact_evidence = capture_failed`, `artifact_capture_error = provider_transport_error`. Metadata history works, but the exact `.3mf` is not archived. A later retest did recover the cached 3MF from the printer over FTPS (with embedded toolpath/G-code renderable in a layer viewer), so the printer *can* serve the artifact — the failure is in the capture path, not a hard capability ceiling.
2. **Duplicate reconciliation is unresolved**: one physical print produced TWO `print_jobs` rows sharing the same external task/project identity (an early metadata-only completed row, then a fuller completed row).

The maintainer reopened the issue (2026-08-23) naming the follow-up scope. This PRD specifies that follow-up concretely against the shipped code.

### 2a. Duplicate: reconciliation keys off one evolving identity

`backend/app/services/printer_hub.py`, `_sync_active_job_db` (lines ~699-836):

- `provider_job_id` is chosen as the FIRST present of `external_task_id` / `external_subtask_id` / `external_project_id` (lines 709-720).
- The dedup lookup (lines 725-733) filters `PrintJob.provider_job_id == provider_job_id`.
- When the resolved row already exists, `provider_job_id` is only *updated* on that row (lines 834-835); it never re-queries under the new id.

Consequence: an early MQTT report carrying only `project_id` creates a row keyed by that project id. A later, richer report whose first present id is now `task_id` (different value) finds no row under `task_id`, so a **second row is created** for the same physical print. That is the two-row duplication observed on the A1 ("early project-ID-keyed completion followed by a richer task-ID-keyed completion").

Secondary contributing factor: the thread also reports the `project_file` request-topic subscription (used to observe capture hints) conflicts with the A1 stability result in issue #69.

### 2b. Capture failure: `provider_transport_error` is a black box

`backend/app/services/printer_hub.py`:

- Capture only triggers when both an artifact path hint is present (`external_artifact_path` from the `project_file` request topic, or `external_gcode_file`) AND `print_stats["_capture_available"] is True` (lines 842-856).
- `_capture_external_artifact` (lines 955-987) calls `download_artifact` → `_download_via_ftps` (`backend/packages/printstash-core/src/printstash_core/printers/bambu.py`, lines 397-449), which wraps the whole FTPS exchange in `except Exception → provider_transport_error` (bambu.py lines 667-681). Any failure — TLS handshake, login, SIZE, RETR, size-mismatch, connection reset — collapses to the single `provider_transport_error` string.
- `_mark_capture_failed` (lines 1036-1048) stores that string in `artifact_capture_error`.

The reporter's own FTPS observation (A1 control-channel can swallow the final STOR reply but still land a complete file) shows the A1 transport is quirky; without a distinct error code we cannot tell a transient transport blip from a path/extension mismatch from an auth failure — so there is no actionable signal for either the code or the user, and no retry story.

## 3. Scope (in)

The follow-up, keyed to the maintainer's reopen note:

1. **Merge evolving identities.** Reconciliation must treat task / subtask / project ids as a related set that maps to one job, not a single preferred id. Design: normalize all three onto the job, and on each incoming report match across all reported ids (any id present on the incoming message that equals any id already stored on an open job for that printer is that same job) before creating a new row. Keep `provider_job_id` as the display-preferred id but never the sole dedup key.
2. **Repair existing duplicates.** A migration/data pass that finds open jobs on the same printer sharing any external task/subtask/project id and merges them into the earliest row, folding the richer metadata (and the later artifact link, if captured) onto the survivor, marking the absorbed rows with the merge reason. Preserve history; never hard-delete.
3. **Regression coverage on the migrated schema.** E2E/integration tests that drive a project-only report followed by a task-only report for one physical print and assert a single row; plus a test for the merge migration against a pre-seeded duplicate pair.
4. **Preserve actionable FTPS failure detail.** Replace the blanket `provider_transport_error` on the artifact path with distinct, auditably-stored codes: auth (`bambu_ftps_auth`), TLS/handshake (`bambu_ftps_tls`), file-missing or path-invalid (`bambu_artifact_not_found` / `bambu_artifact_path_valid`), too-large (`bambu_artifact_too_large`, already present), size-mismatch (`bambu_artifact_size_mismatch`), and connection/reset/timeout (`bambu_ftps_transport`). Surface the code (plus a human one-liner) in `artifact_capture_error` and in the history UI, distinct from `capture_failed`-as-generic. Add a bounded single retry on transport-class codes before falling to metadata-only.
5. **Expose metadata/evidence clearly in history.** The UI must unambiguously distinguish fully reproducible (artifact archived) / partially reproducible (metadata captured, artifact unavailable, error code shown) / basic external job, and never let `subtask_name` read as the archived file (issue #70's "known good print memory" requirement). Where the archived 3MF is available, expose download + embedded toolpath preview surface (the reporter demonstrated the 3MF→embedded-G-code extraction renders real layers).

## 4. Explicitly NOT in scope

- Cloud/Bambu handoff-mode capture: this PRD covers LAN-mode FTPS cache recovery as shipped. Cloud/Studio-account artifact retrieval is a separate feature decision.
- Reconstructing slicer settings the printer never reported. Metadata stays honest: `metadata-only` means exactly that; unknown fields stay blank rather than invented.
- Filament live-consumption for Bambu (printer does not report it; model notes it as null).
- New Bambu hardware support beyond what the provider already declares.

## 5. Acceptance criteria

- One physical A1 external print → exactly one `print_jobs` row after reconciliation, however its MQTT reports arrive (project-only, task-only, or both).
- A pre-existing duplicate pair on one printer merges to a single row with the richer metadata and artifact link kept, and absorbed rows carry a merge-reason marker.
- An FTPS capture failure records a distinct `artifact_capture_error` code (not the blanket `provider_transport_error`) that maps to a user-readable one-liner, and a transient transport failure retries once before resolving `capture_failed`.
- Successful capture stores the exact artifact (SHA-256 verified, as the e2e already asserts) and the history UI labels the job's reproduce-ability state exactly.

## 6. Evidence / sources checked

- `xiao-villamor/PrintStash` clone at tag `v0.12.1` (HEAD `afbe54dd`, 2026-08-20), files cited above by path and line. Retrieved 2026-08-24.
- Issue thread: `#70` (reopened, comments through 2026-08-24), `#69` (Bambu LAN Paho/stability). Retrieved 2026-08-24.
- `CHANGELOG.md` 0.12.0 entry (external-print recorder + best-effort capture + metadata-only labeling). `backend/tests/e2e/test_e2e_bambu_external_capture.py` (shipped acceptance shape: `artifact_evidence = gcode_archived`, sha256 non-zero).

xiao-villamor:
## Summary

- extract framework-neutral backend domain services and utilities while preserving compatibility facades and external API contracts
- construct task-queue, storage, printer-provider, realtime, and session adapters explicitly at application startup while retaining local-first defaults
- extract shared frontend domain utilities and reusable UI primitives into internal workspace packages
- replace server-side authenticated marketplace fetching with browser-extension package capture, including guided setup and localhost access
- migrate the frontend to Oxlint, Oxfmt, and TypeScript 7 with the anti-slop ruleset
- carry forward the urgent 0.12.1 Docker fix and synchronize the branch with current main
- fix dense-STL fallback thumbnails with bounded scanning/raster work, honest partial metadata, and adversarial input limits
- stabilize Bambu A1 MQTT by subscribing persistently only to the report topic while preserving request-topic commands and project-file capture hints
- reconcile evolving Bambu task/subtask/project identities without conflating printers or separate print cycles, and repair historical duplicates without deleting audit rows
- classify FTPS capture failures, retry only explicit transport faults, and expose exact/metadata/basic reproducibility evidence in print history
- archive and safely preview embedded G-code from captured Bambu 3MF projects with RBAC, S3/ZIP resource bounds, localized UI, and accessible viewer controls

## Verification

- integration history and `git diff --check`: pass
- Ruff: pass
- Pyright: 0 errors
- Alembic: one head, empty-database upgrade passes, migration suite 28/28
- focused backend integration: 50/50 plus 3/3 reconciliation regressions
- focused frontend tests: 22/22
- frontend lint, format check, typecheck, and production build: pass
- consolidated review completed with one bounded correction pass; no remaining blocking findings

## Environment limitations / follow-ups

- optional `cascadio` is unavailable in the local integration environment
- some async/provider backend groups hung in this container, and Playwright could not start its `webServer`; CI remains the authoritative full-matrix gate
- non-blocking cleanup remains for an obsolete mock-E2E preview route, absorbed-job filtering in health metrics, and one duplicate `live()` filter

## Scope notes

- replaces #76, which GitHub closed when its head branch was renamed from 0.12.1 to 0.12.2
- supersedes #78; its documentation outcome is already present through current main
- this is the final version-branch integration; future bugs/features return to one short-lived branch and PR per change

Closes #67
Closes #69
Closes #70
Closes #82


xiao-villamor:
Superseded by #86, which contains the complete 0.12.2 history plus the integrated 0.13.0 capture and release-validation work. Issue-closing references were transferred to #86.

PR #86, opened 2026-08-25
feat: consolidate 0.13.0 capture, storage tiers, and Bambu reproducibility

### LINKED ISSUE #67: [Bug]: Thumnails not generated

### LINKED ISSUE #69: [Bug]: Bambu LAN on Bambu A1 fails with Paho v2 callbacks, PUBACK handling, cleanup order, and partial MQTT reports

### LINKED ISSUE #70: [Feature]: Preserve the exact printed artifact/revision and slicer settings for Bambu external print jobs

### LINKED ISSUE #82: PRD: Bambu LAN reproducibility — exact artifact capture + duplicate reconciliation (#70)

### LINKED ISSUE #83: [Bug]: PUID and PGID variables not respected
### What happened?

Seems the Docker containers are ignoring the "PGID" and "PUID" variables.
Having the option to start / execute the container as a specific user could improve the overall security as well as avoiding issues with shared folders permissions (which happened in my case).

### Steps to reproduce

1. Customize the docker-compose configuration by adding the PGID and PUID environment variables
2. Set them properly
3. Start the containers

Image

Image

Image

Pleae note I've tried both compose syntax (_**VARIABLE: ${field}**_ as well as _**- VARIABLE=field**_) the result is the same with either one.

### Area

Install / Docker

### Version or commit

v0.12.1

### Environment

Host OS: Ubuntu 26.04
Install method: Docker
Browser: Edge and Chrome

### Relevant logs

```shell
The above screenshots are taken from Portainer GUI, to be sure, I've tried also a "classic" docker-compose deployment and the behaviour is the same, so it's not Portainer related.
```

### LINKED ISSUE #84: [Bug]: 3mf models generated by 3D Builder are not rendered correctly
### What happened?

OrcaSlicer (and others) as well as 3D Builder by default save projects as .3mf files.

This creates some confusion / issues on PrintStash, as the .3mf file generated by 3D Builder is properly uploaded and added but the 3D render is not working

### Steps to reproduce

1. Generate / edit / remix any STL with 3D Builder and save it with the default options (3mf file).
2. Upload it to PrintStash
3. Go to the uploaded model and open it

The render is not working

### Area

STL / 3MF processing

### Version or commit

v0.12.1

### Environment

Host OS: Ubuntu 26.04
Install method: Docker
Browser: Edge and Chrome

### Relevant logs

```shell

```

### LINKED ISSUE #88: [Feature]: Support GFM tables in markdown documents
### What

Tables written in a document's Markdown are not rendered as tables. The raw pipe syntax shows up as literal text instead.

Tested on 0.12.1 (self-hosted, `printstash-api-lite` + `printstash-frontend`)

### Why this matters for a 3D model catalogue

Print documentation is table-shaped more often than not. The things people paste into a model's document are almost always a small matrix:

| Part    | Filament | Nozzle | Layer  | Infill |
|---------|----------|--------|--------|--------|
| Body    | PETG     | 0.4    | 0.2 mm | 25 %   |
| Hinge   | PLA      | 0.4    | 0.12mm | 60 %   |

Print settings per part, filament per colour, a BOM of screws and inserts, post-processing steps. That content also arrives pre-formatted: READMEs on Printables, MakerWorld and Thingiverse are GitHub-flavoured Markdown, so a copy-paste into PrintStash silently loses its structure and turns into a wall of pipe characters.

### Cause

`frontend/package.json` pulls in `react-markdown ^10.1.0`, but no `remark-gfm`. Plain react-markdown implements CommonMark only, and tables are a GFM extension, so they are correctly (per CommonMark) rendered as a paragraph

### Proposed solution

Add `remark-gfm` and hand it to the existing `<ReactMarkdown>` in the document viewer:

```jsx
import remarkGfm from 'remark-gfm'

<ReactMarkdown
  remarkPlugins={[remarkGfm]}
  rehypePlugins={[rehypeSanitize]}>
  {content}
</ReactMarkdown>
```

### Area

File organization

### Alternatives considered

_No response_

### LINKED ISSUE #90: [Feature]: broad storage backend support via OpenDAL, behind explicit capability tiers
## Problem

Self-hosters keep asking PrintStash to store their library somewhere it does not
speak: Nextcloud over WebDAV, a NAS over SFTP, occasionally a consumer drive.
Today the answer is local filesystem or S3-compatible, and nothing else.

S3-compatible already covers more than it looks like — B2, R2, MinIO, Wasabi,
Ceph, Garage all work through it. What it does not cover is the setup a large
share of self-hosters already run: **a Synology or a Nextcloud box that is
already the household's file server.** Asking those users to stand up MinIO
next to storage they already have is a real adoption cost, and writing a bespoke
adapter per protocol is not tractable for a project this size.

### What those users do today, and why it is worse than it looks

They mount the remote share into the container and point `VAULT_DATA_DIR` at it
— NFS, SMB/CIFS, `sshfs`, `rclone mount`, `davfs2` for WebDAV — and run the
local backend on top. It appears to work. **It is silently outside the
guarantees the local backend claims**, because that backend is built on POSIX
primitives network and FUSE filesystems do not reliably provide:

- `os.link(temp, dest, follow_symlinks=False)` is our atomic no-replace
  publication. SMB/CIFS, `davfs2`, and `rclone mount` generally do not support
  hardlinks at all, so `create_stream` either fails outright (the safe outcome)
  or behaves unlike the contract on mounts where it half-works.
- `_fsync_directory` opens the parent with `O_DIRECTORY` and `fsync`s it. On
  many FUSE and network filesystems that is a no-op, so the durability
  guarantee quietly evaporates.
- **`CreationReceipt` fingerprints `(st_dev, st_ino, st_ctime_ns, st_size)`.**
  FUSE layers synthesize inode numbers and several reassign them per mount
  session. Receipts persist (`CaptureUploadSlot.receipt_json`), so after a
  container restart a stored receipt can no longer match the object it
  describes — `creation_matches` returns `False` for the rest of that object's
  life, and rollback and verification paths fail closed permanently.

And nothing detects this: `LocalStorageBackend.ensure_setup()` does two
`mkdir`s. It never probes whether the filesystem beneath it can honour any of
the above, and `docs/known-limitations.md` does not warn about it. PrintStash
reports the backend as `local` and behaves as though every guarantee holds.

## What we want to achieve

Broad storage backend support — WebDAV, SFTP/FTP, Azure, GCS, the consumer
drives — **without quietly weakening the guarantees PrintStash makes about a
user's files.** Every backend is labelled with what it does and does not
promise, and the ones that promise less require the operator to say so out loud.

## Why OpenDAL

[Apache OpenDAL](https://opendal.apache.org/) is a Rust data-access layer with
Python bindings that speaks ~60 storage services behind one interface. One
adapter, written once, reaches all of them. The alternative — an adapter per
protocol, each with its own auth, retry, and streaming behaviour — is work we
would never finish and could not keep tested.

What it is **not** is a replacement for what we have. See "What we deliberately
are not doing".

## What it buys, concretely

We checked which services can actually uphold PrintStash's write contract.
`write_with_if_not_exists` is declared by exactly **fs, s3, gcs, azblob, azdls,
oss, cos, ghac** — that is the entire list.

| | Backends |
|---|---|
| Conditional create **and** user metadata | `fs`, `s3` *(both already implemented by hand)*, `gcs`, `azblob`, `azdls`, `oss`, `cos` |
| Neither | `webdav`, `sftp`, `ftp`, `onedrive`, `googledrive`, `dropbox`, `pcloud`, `koofr`, `yandexdisk`, … |

This is the crux of the whole proposal, and it is worth stating bluntly: **the
backends that arrive with full guarantees are the three hyperscaler object
stores, and the backends people actually ask for are the ones that cannot make
those guarantees.** A version of this work that supports only the safe row
delivers Azure and GCS to an audience that mostly wants Nextcloud. The support
question and the safety question are the same question.

So the proposal is to support both, and to make the difference explicit,
enforced, and honest — rather than either refusing the second row or shipping it
pretending nothing changed.

## The tier taxonomy

Backends are not "safe" and "unsafe". PrintStash's write contract needs several
separable things, and they fail independently — so tiers are **derived** from
capability axes rather than assigned by hand:

- `conditional_create` — can a create fail rather than clobber?
- `object_identity` — can a receipt bind to *these exact bytes*? (inode /
  version / etag / none)
- `verified_delete` — can we delete *only* the proven object?
- `conditional_replace` — replace only while the proof still holds?
- `namespace_ownership` — can we prove a key sits inside our own root?
- `direct_path` — is a real filesystem path available? (performance, not safety)

Which yields three tiers, each named for what it **guarantees**:

| Tier | Guarantee | Backends |
|---|---|---|
| **Verified** | A create never clobbers; a rollback destroys only our own bytes | local, versioned S3, versioned GCS/azblob |
| **Guarded** | A create never clobbers; a failed operation leaks bytes we cannot positively reclaim | **unversioned S3 — shipping today**, unversioned GCS/azblob |
| **Unguarded** | Best effort; concurrent writes to one key can lose data | webdav, sftp, ftp, consumer drives |

**The middle tier already ships.** `S3StorageBackend.rollback_create` already
returns `False` and logs *"S3 object has no immutable version identity"* when
the bucket is unversioned. We are not inventing a taxonomy — we are naming one
PrintStash has had since S3 support landed, and currently reports to nobody.

Two useful consequences of that:

1. The tier machinery is **buildable and shippable before any OpenDAL code
   exists**, tested against unversioned MinIO.
2. Existing S3 users get honest reporting about their own setup as a side
   effect.

## How we keep it safe

Four mechanisms, in the order they engage:

1. **A DB intent ledger.** Insert the storage key *before* writing bytes; the
   unique constraint becomes the conditional-create primitive for backends that
   have none. One API process is the supported topology (`known-limitations.md`,
   and startup claims a vault lock), so the concurrency this closes is
   *within-process* — two overlapping uploads served by the same event loop,
   which is the realistic case. It narrows the honest warning from "atomicity is
   unavailable" to "an external tool writing into the same folder can still
   collide".
   PrintStash already built this shape once — `CaptureUploadSlot` has `state`,
   a unique `storage_key`, `sha256`, and `receipt_json`.
2. **An orphan sweep.** `PENDING` ledger rows past a threshold are reclaimed.
   This recovers leaked bytes on **every** tier including unversioned S3 today,
   with no OpenDAL involved.
3. **A boot gate.** An Unguarded backend refuses to start without
   `VAULT_STORAGE_ALLOW_UNVERIFIED=true`. Environment-only, never a UI toggle —
   a runtime warning is read by nobody, and the flag gates the very boot that
   would serve the UI. The UI's job is to *display* the tier and grey out
   unverified options, not to flip the acknowledgement.
4. **A separate delete gate.** Hard delete requires explicit confirmation on any
   non-Verified backend even after the flag is set. The blanket acknowledgement
   covers ingest; it does not cover irreversible deletion.

Warnings are generated per absent axis, not per tier, so the operator is told
the failure they will actually meet ("two simultaneous uploads of the same
revision can silently overwrite each other") rather than a category label.

**And a fifth that applies to a backend we already ship:** `LocalStorageBackend`
must *probe* its capabilities at `ensure_setup()` rather than assume them —
attempt a hardlink in the data dir, attempt an `O_DIRECTORY` fsync, and detect a
network or FUSE filesystem (`statfs` `f_type`, `/proc/mounts`) whose inode
numbers may not survive a remount. A mounted-share install then reports Guarded
or Unguarded honestly instead of claiming Verified. This needs no OpenDAL, helps
users who are on a mount **today**, and is the single highest-value item in the
whole proposal for existing installs.

## Users configure providers, never OpenDAL

OpenDAL is an implementation detail and must not appear in the product. No
`VAULT_OPENDAL_*` variable, no scheme string, no options blob — a self-hoster
picks **Nextcloud** and fills in a Nextcloud form.

So every backend is modelled as a first-class typed provider, **including the
two that already exist**:

```
StorageProvider     what the operator picks
                    local · s3 · nextcloud · webdav · sftp · azure_blob · gcs
      ↓             typed config model + UI metadata + expected tier
TransportSpec       how we actually talk to it  (internal, never rendered)
                    NATIVE_LOCAL · NATIVE_S3 · OPENDAL(scheme, root, options)
      ↓
StorageBackend      the existing abstraction — unchanged
```

The middle layer earns its place because the mapping is **many-to-one**:
Nextcloud, ownCloud, Seafile, Synology and generic WebDAV all resolve to one
WebDAV transport, and R2/B2/Wasabi/MinIO all resolve to the S3 backend we
already have. Without that seam the fan-in would have to live inside the OpenDAL
adapter — which would then need to know what Nextcloud is — or be duplicated
across every provider. It also means a provider can switch transport later
without any operator's configuration changing.

Concretely, the Nextcloud provider asks for a URL, a username, an app password,
and a folder. It knows on the user's behalf that Nextcloud serves WebDAV at
`/remote.php/dav/files/{username}` — under a raw-scheme configuration the
operator would have to know that themselves. **A provider is product knowledge;
a scheme is library trivia.**

Configuration looks like this, and nothing in it names the library:

```
VAULT_STORAGE_PROVIDER=nextcloud
VAULT_NEXTCLOUD_URL=https://cloud.example.com
VAULT_NEXTCLOUD_USERNAME=printstash
VAULT_NEXTCLOUD_APP_PASSWORD=…
VAULT_NEXTCLOUD_FOLDER=/PrintStash
```

### The providers we intend to ship

Grouped by how we talk to them, because that determines both cost and tier.
"Native knowledge" is what the provider encodes so the operator doesn't have to.

| Provider | Talks via | Native knowledge it encodes | Expected tier |
|---|---|---|---|
| Local filesystem | native | — | Verified *(probed — a mounted share may be lower)* |
| S3-compatible | native | — | Verified / Guarded by bucket versioning |
| **Cloudflare R2** | native S3 | endpoint `https://{account_id}.r2.cloudflarestorage.com`, `region=auto` | as S3 |
| **Backblaze B2** | native S3 | S3 endpoint per region | as S3 |
| **Wasabi** | native S3 | regional endpoint | as S3 |
| **MinIO / Garage / SeaweedFS** | native S3 | path-style addressing, self-hosted defaults | as S3 |
| **Nextcloud** | WebDAV | `/remote.php/dav/files/{username}`, app-password guidance | Unguarded |
| ownCloud | WebDAV | `/remote.php/dav/files/{username}` | Unguarded |
| Seafile | WebDAV | `/seafdav` | Unguarded |
| Synology DSM | WebDAV | WebDAV Server package, ports 5005/5006 | Unguarded |
| Hetzner Storage Box | WebDAV / SFTP | documented host and path layout | Unguarded |
| Generic WebDAV | WebDAV | — (typed escape valve *within* the family) | Unguarded |
| NAS over SFTP | SFTP | host/port/user/key-or-password/path | Unguarded |
| Azure Blob Storage | azblob | container + account naming | Guarded |
| Google Cloud Storage | gcs | bucket + service-account JSON | Guarded |

**The four bolded S3 rows are the cheapest value in this whole proposal.** R2,
B2, Wasabi and MinIO become named providers that preset the endpoint and region
over the S3 backend we *already ship* — no new transport, no new adapter, no
OpenDAL, and they remove the single most common S3 setup mistake, a hand-typed
endpoint. Same trick as Nextcloud, applied to something that already works. It
lands in the catalogue phase, before any OpenDAL code exists.

Deliberately **out of the first cut**: Dropbox, Google Drive, OneDrive. Not for
lack of a transport — each needs an OAuth authorization flow with token storage
and refresh, which is a larger change than this entire proposal and unrelated to
storage semantics. Candidates once a general OAuth credential story exists.

Two open questions for the catalogue phase: how SFTP key material is supplied (a
mounted key file, or an inline key encrypted at rest?), and whether vendor rows
that encode *nothing* deserve to exist — Synology-over-SFTP is generic SFTP with
a different label, and a vendor entry earns its row only when it removes real
knowledge from the operator the way Nextcloud does.

Two further consequences worth calling out:

- **One declaration, five surfaces.** Env var names, the runtime-config overlay,
  the Settings form, the docs table, and the tier badge are all generated from
  the provider's spec. They currently drift; after this they cannot. The Settings
  form is generated from the pydantic model, so adding a provider needs **no
  frontend change**.
- **No generic passthrough.** There is deliberately no way to point PrintStash at
  an arbitrary OpenDAL scheme. An untested scheme has no `contract/` coverage and
  therefore no tier we could state truthfully — and the tier claim is the whole
  deliverable. Adding a provider is a small mechanical PR (a model, a
  `transport()`, a registry entry, a contract test), which is cheap enough that
  the hatch is unnecessary.

`VAULT_STORAGE_BACKEND=local|s3` keeps working as a deprecated alias, so no
existing `.env` breaks.

## What the setup experience looks like

A flat provider dropdown is already ~15 rows and grows. Worse, it makes the
operator pick blind and discover the safety consequence only after filling in a
form. So the picker is **two-step, grouped by transport family** — which happens
to be the same axis that determines tier, so navigating it teaches the guarantee
story rather than springing a badge at the end:

```
Step 1 — category (5 cards, no scrolling)
  ┌──────────────┬──────────────┬──────────────┬──────────────┬──────────────┐
  │ This machine │ S3-compatible│  Nextcloud   │   NAS over   │    Cloud     │
  │              │object storage│  & WebDAV    │     SFTP     │  providers   │
  │  ✓ Verified  │  ✓ Verified  │ ⚠ Unguarded  │ ⚠ Unguarded  │  ~ Guarded   │
  └──────────────┴──────────────┴──────────────┴──────────────┴──────────────┘

Step 2 — provider within that family (2–6 rows, each with its tier badge)
Step 3 — the generated form for that provider
```

The tier badge shows at **step 2, before** anything is typed. "Generic WebDAV"
sits last inside its own family as the typed escape valve.

This composes existing `components/ui/` primitives only (`card`, `tabs`, `badge`,
`input`, `modal`) so it needs **no new primitive** — relevant given `DESIGN.md`'s
compose-don't-hand-roll rule. A searchable combobox would have to be built from
scratch; a category grid doesn't.

**The form is described by the server**, via `GET /api/v1/storage/providers`
returning label, category, blurb, expected tier, consequences, docs link, and a
field descriptor list (label, help text, type, secret-ness, required-ness)
projected straight from the pydantic model. So **adding a provider is a
backend-only change** — no frontend edit, no new route. Secret fields render
masked and are write-only; the API never returns a stored secret, only whether
one is set.

## Documentation

Five places, distinct audiences — and one naming problem to fix first.

**`docs/provider-support.md` is titled "Printer Provider Support."** Adding
storage providers overloads a word `CONTEXT.md` already flags as ambiguous for
"Backend". So unqualified "provider" keeps meaning the printer kind, and storage
code says "storage provider" in full.

| Where | Audience | Content |
|---|---|---|
| **`docs/storage-providers.md`** *(new)* | **operators deploying** | one section per storage provider: env vars, worked example, its tier and what that implies, setup gotchas |
| `CONTEXT.md` | contributors | binding vocabulary: Storage provider, Transport, Verified/Guarded/Unguarded |
| `docs/known-limitations.md` | operators diagnosing | the network-mount warning **currently absent**, plus per-tier caveats |
| `UPGRADE.md` | upgraders | the `VAULT_STORAGE_BACKEND` alias; the dropped `s3:CreateBucket` / `s3:PutLifecycleConfiguration` grants |
| `.env.example` | first-time setup | the new variables, commented |

The provider table in `docs/storage-providers.md` is **generated from the
registry**, with a `repo/` test asserting the doc matches it — same anti-drift
pattern as `test_openapi_contract.py`. Documentation becomes the sixth consumer
of the one declaration, and the only one a human could otherwise forget.

## External libraries: local-fs transport stays; the catalogue reaches them in #92

Two questions hide in "should external libraries get provider support", with
different answers.

**Read bytes through a remote transport instead of a mount?** Not on the strength
of anything in this issue — the argument justifying everything above does not
transfer, and one that runs the other way does (below).

**Be *configured* through the same provider catalogue?** Yes, eventually, on
**coherence** grounds that none of the below touches: once the vault is
configured as "Nextcloud, this URL, this app password", an external library still
demanding `/mnt/nextcloud/PrintStash` leaves one product with two ways to name a
remote storage location. That is a configuration cost, not a correctness one, and
it is tracked in **#92** — scoped read-only, write-back excluded.

The case against the *transport* migration, which is what this issue settles:

**The safety argument doesn't apply.** Everything hazardous about running the
*vault* on a mounted share is write-side — `os.link` publication, directory
`fsync`, inode fingerprints, receipt verification, quarantined rollback. External
libraries do none of those to source bytes. Per `CONTEXT.md` the folder is the
source of truth, PrintStash never overwrites or deletes a linked file's bytes,
and trash/GC skip external blobs. `get_backend()` appears in
`external_library.py` exactly once, for a derived thumbnail that lands in vault
storage. So for a read-only external library a mount isn't a dangerous
workaround — it's just how you read a remote folder. Nothing is being silently
downgraded because nothing is being guaranteed.

**A mount is strictly better here.** `watchfiles` gets inotify events on a local
filesystem; WebDAV and SFTP have no change-notification mechanism at all. Moving
external libraries onto providers would *remove* a capability.

**And there's no performance case.** Indexing a remote library downloads every
file once regardless — hash and thumbnail both need full bytes, and
`_reindex_changed` re-hashes on any size/mtime change. FUSE and OpenDAL push the
same bytes over the same wire.

### External libraries are the model to copy, not the thing to fix

This subsystem already does what phase 1 asks the vault to start doing:

| | External libraries | Vault |
|---|---|---|
| Detects filesystem kind | ✅ `detect_fs_kind()` | ❌ |
| Persists it | ✅ `ExternalLibrary.fs_kind` | ❌ |
| Exposes it | ✅ API returns `fs_kind` + `watch_active` | ❌ |
| Degrades honestly on it | ✅ `should_watch()` declines `AUTO` off-local | ❌ |

Detect, persist, expose, degrade — four for four, versus zero. **The borrowing
direction is external-library → vault**, which is exactly what phase 1 does by
promoting `detect_fs_kind` to a shared module.

### The follow-up, and what it must confront

**#92** carries this. It is justified by setup ergonomics and product coherence,
**not** by safety or throughput, and should be reviewed on those terms.

It also has to confront the one place mounting is genuinely superior: real-time
watching works over a mount and can never work over a transport, so a
provider-backed library is **scheduled-scan-only by construction** and needs a
resting state distinct from "network mount, might work someday". `EVENTS` has to
be rejected at validation rather than silently doing nothing.

A sharper trigger exists for the transport itself: an operator who **cannot**
mount — a managed container platform forbidding FUSE or privileged containers,
with files in a hosted Nextcloud. Narrow but real.

The catalogue is therefore defined **storage-role-agnostic**, so #92 needs no
decision here reopened.

## What we deliberately are not doing

**Local never moves to OpenDAL.** OpenDAL's API is the intersection of ~60
backends — path-addressed, because that is what they all share. PrintStash's
local guarantee is *stronger* than the object-store guarantee: `os.link`
no-replace publication, directory `fsync`, and an inode fingerprint that lets
rollback destroy only bytes it can prove it wrote. Routing local through a
uniform interface would silently downgrade the **default** backend that every
self-hoster runs, and it matters most there, because a self-hoster's
`/data/files` is a shared, multi-writer directory in a way an S3 prefix is not.

We want the **union** of guarantees, where each adapter contributes its
strongest primitive — not the intersection. The full argument, with the specific
TOCTOU window at stake, is in the ADR.

**S3 stays native on boto3 — decided, not deferred.** Measured against the
published wheels rather than the Rust documentation: `Operator.write()` returns
`None` in `opendal` 0.47.6, so an OpenDAL S3 create yields no `version_id`, and
a follow-up `stat()` is racy by construction (it reports whichever version is
current, possibly someone else's). An OpenDAL-backed S3 adapter therefore caps
at **Guarded**, and migrating would **demote a backend that is Verified today**
— in exchange for +17 MB of image and a faster import nobody would notice.
Revisiting requires a specific upstream change (the Python binding returning
write metadata with `version_id`); until then there is nothing to weigh.

boto3 also does not leave the tree either way: `services/backup.py` runs its own
client against a different bucket with different credentials, and that stays
regardless of anything decided about storage.

## Impact

**For users**

- **No more FUSE or network mounts to reach a NAS or Nextcloud.** Speak WebDAV
  or SFTP directly instead of `davfs2`/`rclone mount`/CIFS underneath the local
  backend. This removes a whole class of setup work (mount units, credentials in
  `fstab`, `--cap-add SYS_ADMIN` or privileged containers for FUSE inside
  Docker, mounts that come back after a reboot in the wrong order) — and, more
  importantly, replaces a configuration whose guarantees are *assumed* with one
  whose guarantees are *measured and stated*.
- Note the honest shape of that trade: over a mount, PrintStash reports Verified
  and cannot actually deliver it. Over OpenDAL WebDAV it reports Unguarded and
  delivers exactly that, plus the ledger's conditional-create. **Correctly
  labelled Unguarded is a stronger real position than mislabelled Verified** —
  and it is the labelling, not the protocol, that is the actual product here.
- Point PrintStash at an existing Nextcloud or NAS instead of standing up object
  storage next to it.
- Azure and GCS as first-class targets for anyone already on those clouds.
- Existing S3 users learn whether their bucket is actually Verified or Guarded —
  a question they cannot currently ask.
- **Existing users on a network mount get told.** The local capability probe
  (below) detects it and reports the real tier instead of assuming POSIX.

**For the project**

- One adapter covers the entire long tail; the per-protocol adapter treadmill
  never starts.
- The storage layer gains a vocabulary for degraded guarantees, which the cloud
  seam (hard rule 5) will need regardless of OpenDAL.
- The orphan sweep and the ledger are net safety wins on the backends we already
  ship.

**Costs, stated honestly**

- A Rust native extension enters the wheel set: **`opendal` is 39.4 MB installed**
  (one 38.6 MB `.so`, every service compiled in, not trimmable) against
  **22.7 MB for `boto3` + `botocore`**. Mitigated by making `opendal` an optional
  extra — a self-hoster on local or S3 never installs it. Wheel coverage is not a
  concern: prebuilt `cp311-abi3` wheels exist for manylinux x86_64/aarch64/armv7l,
  musllinux, macOS x86_64/arm64 and Windows, so both published image
  architectures and local dev install without a Rust toolchain.
- Three adapters to maintain instead of two.
- PrintStash will ship a configuration in which concurrent writes can lose data.
  That is the deliberate trade, bounded by the five mechanisms above.

## Risks

- **A declared capability can be wrong.** `write_with_if_not_exists` has shipped
  as declared-but-unenforced, reported against azblob *and* s3/MinIO. Mitigation:
  the `contract/` tier must prove conditional create against a real server per
  scheme, and the adapter pins an exact `opendal` version. This is not optional
  coverage.
- **Upstream's write-returns-metadata work is still landing per service.** A
  scheme that does not populate `version_id` lands in Guarded rather than
  Verified — degradation by design, but it means a scheme's tier can change
  across upstream releases and must be reported at runtime, not documented as a
  constant.
- **Ledger drift** could point a sweep at a live object. Mitigated by
  insert-then-write ordering, re-verification before any sweep delete, and
  restricting the sweep to never-committed rows.

## Proposed sequencing

Phases 1–3 are valuable on their own and touch no new dependency. If the OpenDAL
half were dropped entirely, they would still be worth shipping.

1. **Capability axes and derived tiers, including the local filesystem probe.**
   `StorageCapabilities` computed at `ensure_setup()`; `LocalStorageBackend`
   probes hardlink support, directory `fsync`, and filesystem kind rather than
   assuming POSIX — **reusing `external_library.detect_fs_kind()`**, which already
   reads `/proc/self/mountinfo` and already treats fuse as unknown, written for
   the analogous reason on the watch side. Tier surfaced in `/health`, Settings,
   and the boot log. Tested against unversioned MinIO and a non-hardlink
   filesystem. *No new dependency, no new detection machinery, and the phase that
   helps the most existing installs — anyone on a mounted NAS share learns their
   real tier.*
2. **The intent ledger and orphan sweep.** Generalise the `CaptureUploadSlot`
   shape; insert-then-write at the ingest seam; a sweeper for never-committed
   rows. *No new dependency; immediate benefit to existing S3 users.*
3. **Retire the bucket-administration writes.** Remove
   `_apply_lifecycle_policy` (PrintStash should not be installing expiration
   rules on the user's bucket — it contradicts our own
   `destructive_lifecycle_findings` guard) and the create half of
   `_ensure_bucket` (drops `s3:CreateBucket` and `s3:PutLifecycleConfiguration`
   from the credential we ask self-hosters for). Documented as a breaking-ish
   operational change in `UPGRADE.md`.
4. **The boot gate and delete gate.** `VAULT_STORAGE_ALLOW_UNVERIFIED`, the
   per-axis warning table, the UI display and greying-out.
5. **The provider catalogue.** `StorageProvider` + typed `ProviderConfig` +
   internal `TransportSpec`; `local` and `s3` migrate onto it as providers so the
   catalogue is uniform; env vars, the overlay, and the Settings form all
   generated from the spec. `VAULT_STORAGE_BACKEND` kept as a deprecated alias.
6. **The OpenDAL transport.** One adapter behind the catalogue, capability probe
   at setup, `contract/` suite per supported scheme. Ships with **Nextcloud**,
   generic **WebDAV**, and **SFTP** as the headline providers; Azure and GCS are
   nearly free once the transport exists.
7. **The picker and the docs.** `GET /api/v1/storage/providers`, the
   category-first picker, `docs/storage-providers.md` generated from the registry
   with its `repo/` guard, the `CONTEXT.md` vocabulary entries, and the missing
   network-mount warning in `known-limitations.md`. Phase 5 can ship a plain
   generated form; this is the phase that makes it pleasant.

Each phase is its own PR per hard rule 4.

## Design and decision record

The low-level design — the capability dataclass, tier derivation, ledger schema,
the provider catalogue and its `ProviderConfig`/`TransportSpec` layering, the
per-axis warning table, and the full intersection-vs-union argument for why local
stays hand-rolled — is in
**ADR-0003**, proposed in **#91** (`docs/adr/0003-storage-capability-tiers.md`).

## Open questions

- Which schemes ship in the first cut? WebDAV and SFTP cover the stated demand;
  Azure/GCS are nearly free once the adapter exists.
- Is `opendal` an optional extra or a base dependency? Optional keeps the
  local/S3 install unchanged, at the cost of a second install path to document
  and test.
- Does the orphan sweep run on the existing periodic loop in `main.py`, or does
  it want its own cadence?

### LINKED ISSUE #110: [Bug]: Setting model collection to None does not save.
### What happened?

In Edit Model, the Collection dropdown includes None, but selecting None and saving does not clear the collection assignment. The model remains in its previous collection.

Moving a model out of a collection is possible via drag-and-drop using the left-hand file tree, so the application does support an uncollected/root state. The bug appears to be specifically in the Edit Model save path for None.

### Steps to reproduce

assign model to collection → Edit → Collection → None → Save → model remains assigned. Meanwhile, drag-and-drop via the file tree can return it to root, proving the root/uncollected state itself works.

### Area

Web UI

### Version or commit

_No response_

### Environment

_No response_

### Relevant logs

```shell

```

xiao-villamor:
## Summary

- Adds runtime-probed Verified, Guarded, and Unguarded storage tiers; durable publication intents; provider-safe recovery; and guarded destructive operations for local, S3-compatible, WebDAV, SFTP, and Google Drive storage.
- Adds read-only Library sources with restart-safe discovery, root binding, source verification, write-back guardrails, durable GC plans, and a reusable Remote storage settings surface for Library sources and backup replicas.
- Adds remote backup destinations for S3, WebDAV, SFTP, and Google Drive. Restore verifies provider identity and SHA-256, while providers without immutable delete identity stay out of automatic retention and Vault-GC evidence.
- Adds Multipart Models with named parts, alternatives, collection and tag editing, guides, local or external covers, favorites, a dedicated workspace, and an Organized library view that keeps reusable member Models discoverable.
- Improves browser capture for Printables, MakerWorld, and Thingiverse. Captures keep per-Artifact provenance, support browser-transferred files and selective retry, and retain provider-specific recovery guidance without receiving marketplace cookies.
- Completes Bambu LAN reproducibility work with stable report reconciliation, bounded FTPS capture, exact, metadata, or basic evidence, embedded 3MF G-code extraction, and archived-toolpath previews.
- Adds BGCODE uploads across browser upload surfaces, deterministic and bounded thumbnail rendering, embedded 3MF thumbnail preference, transformed 3MF previews, GitHub-flavoured Markdown tables, direct Model tag editing, and a settings restart action for supervised deployments.
- Fixes Pending Import lifecycle races, provider file discovery, cached preview loading, theme persistence under CSP, upload request headroom, long backup-restore details, legacy storage upgrades, and clearing a Model Collection from Edit Model.
- Extracts reusable backend logic into `printstash-core`, shared frontend domain and UI packages, explicit local-first runtime composition, mirrored test tiers, coverage ratchets, isolated Playwright suites, Docker pre-release coverage, and native arm64 builds.

## Testing

- [ ] Backend: the current head has a pending full backend gate. The Python 3.13 advisory compatibility check is currently failing and needs follow-up.
- [ ] Frontend: the current head has a failing frontend check and needs follow-up.
- [x] Browser extension, API image builds for amd64 and arm64, MinIO-to-SeaweedFS migration, printer-core dependency matrices, and the SQLite async-extra check currently pass.
- [x] Focused issue #110 check: `pnpm exec vitest run src/components/model-detail/__tests__/index.test.tsx` passed with 37 tests. Lint, format, and typecheck also passed.

## Coverage matrix

| # | Behaviour | Category | Precondition / input | Observable outcome | Tier | Status |
|---|-----------|----------|----------------------|--------------------|------|--------|
| 1 | Storage capability probes classify provider guarantees | Happy / edge / error | Local and remote storage configurations | Runtime capability reports and destructive-operation guardrails match observed guarantees | Unit / integration / contract | ✅ |
| 2 | Library-source discovery remains bounded and restart-safe | Happy / edge / error | Mounted and remote source scans, changed roots, unavailable providers | Imports, tombstones, source verification, and removals preserve source ownership boundaries | Integration / E2E | ✅ |
| 3 | Remote backups preserve provider identity | Happy / edge / error | S3, WebDAV, SFTP, or Google Drive backup destinations | Replication and restore validate declared provider and archive digest | Unit / integration / contract | ✅ |
| 4 | Multipart Models preserve independent member ownership | Happy / edge | Parts, alternatives, covers, guides, collections, tags, and favorites | Grouping changes do not move or delete member Models or Artifacts | Integration / frontend unit / Playwright | ✅ |
| 5 | Browser capture retains accurate provenance and outcomes | Happy / edge / error | Browser-transferred files and mixed provider responses | Complete and partial outcomes, safe provider recovery, and source snapshots persist across retry and refresh | Backend E2E / extension / Playwright | ✅ |
| 6 | Bambu print evidence stays reproducible | Happy / edge | External print reports with exact or partial evidence | Stable job identity, evidence level, and archived toolpath remain available | Backend E2E / frontend unit | ✅ |
| 7 | Preview and upload formats remain bounded and supported | Boundary / error | Dense STL, embedded 3MF preview, transformed 3MF, and BGCODE uploads | Safe thumbnail fallback, correct transformed preview, and accepted binary G-code uploads | Unit / integration / frontend unit | ✅ |
| 8 | Model editing can clear its Collection | Regression | Model assigned to a Collection, then Edit Model selects None | PATCH sends the root sentinel and the assignment is removed | Frontend unit | ✅ `src/components/model-detail/__tests__/index.test.tsx` |
| 9 | Test infrastructure enforces mirrored ownership and coverage floors | Repository rule | Full backend, frontend, extension, and Playwright suites | Test tiers, factories, hygiene, and ratchets stay enforceable in CI | Repository / CI | ✅ |

## Notes

- **Back up before upgrading.** This branch contains additive capture, storage-publication, provider-configuration, multipart, remote-backup, and schema-convergence migrations. Review [UPGRADE.md](https://github.com/xiao-villamor/PrintStash/blob/0.13.0/UPGRADE.md) before deployment.
- Existing S3-compatible buckets must be provisioned before startup. PrintStash inspects them for safety diagnostics but does not create buckets or change lifecycle policies. Unguarded storage requires `VAULT_STORAGE_ALLOW_UNVERIFIED=true`.
- Local and native S3 remain dependency-free defaults. WebDAV, SFTP, and Google Drive transports are available in the full API image, not the lite image.
- Provider secrets are write-only in API responses. Legacy storage environment variables remain compatibility inputs during upgrade.
- Bambu LAN remains beta. Metadata-only history is expected when the printer cache no longer exposes the exact bytes.
- Version bump, tag, images, and GitHub release publication remain separate release steps.

Includes integrated work from #80, #85, #89, #90, #91, #93, #94, #95, #96, #97, #99, and #100.

Closes #67
Closes #69
Closes #70
Closes #82
Closes #83
Closes #84
Closes #88
Closes #90
Fixes #110


jorgehermo9:
Closes #90 #92

PR #89, opened 2026-08-27
test: full test infrastructure overhaul

jorgehermo9:
## Summary

- 

## Testing

- [ ] Backend tests added/updated, or not needed
- [ ] Frontend checks run, or not needed
- [ ] Manual test notes included, or not needed

## Notes

Mention any schema, API, storage, or printer-provider behavior changes.


jorgehermo9:
@xiao-villamor this is ready for review. Below is everything the overhaul turned up, in one place — what changed, what it found, and what is genuinely still open.

## Where the suite ended up

| | Branch base | Now |
| --- | --- | --- |
| backend suite | 2 733 passing, **10 failing** | **6 035 passing, 0 failing, 0 skipped** |
| `app/` coverage | 91.25 %, lines only | **93.62 % with branches on**, per-module floors |
| `printstash-core` suite | 342 tests, never gated in CI | **1 396 tests**, 7 s, floor at 98.8 % |
| frontend vitest | 572 tests, `src/lib/**` only | **1 447 tests** across app + both packages |
| `@printstash/ui` coverage | 32.36 % / 15.95 % | **99.17 % / 98.15 %**, 100 % of lines |
| `@printstash/domain` coverage | 88.38 % / 82.10 % | **95.48 % / 92.63 %**, 100 % of lines |
| app coverage | 86 % of 19 % of the code | **80.96 % / 73.43 %** of all of it, gated |
| Playwright mock-API | 30/32 (2 red for months) | **32/32** |
| skipped tests | 16 | **0**, and a skip now fails |
| test names joining two behaviours | 313 | **0** (backend), capped and falling (frontend) |
| files with a contract header | 162/201 | **all of them**, enforced |
| cross-module test imports | 4 | **0**, enforced |

Coverage is measured with **branches on** in all three suites and gated **per module**
rather than blended, so a module cannot hide behind the average. Every floor is
two-sided: fall below it and the run fails, rise clear of it and the run *also* fails
until the floor is raised. That is why many commits here edit a floor — that edit is
the point.

## Production bugs found and fixed (21)

The suite found these; none were reported by a user.

**Live 500s and permanent feature failures**

- **B-4** — two live 500s from comparing a naive and an aware `datetime`, on paths a
  self-hoster hits routinely. Both had been shipping.
- **B-10** — Bambu artifact capture refused every URL the printer actually reports, so
  capture for an externally started print had *never once worked*, on a feature
  documented as supported.
- **B-11** — a bare-array file listing crashed the printer poll loop.
- **B-12** — the Moonraker websocket loop could spin without yielding, starving the API
  process's event loop. Latent, and only visible on 3.13.
- **F-8** — `POST /libraries/{id}/scan` returned 500 whenever a scan was already
  running.

**Found by turning foreign keys on** (the suite had them *off* for months — see below)

- **B-16** — `DELETE /libraries/{id}` trashed the indexed files and then deleted the
  library those files still pointed at.
- **B-17** — `hard_delete_file` cleaned 3 of the 5 NOT NULL children of `files`, so
  purging a file that was in a print batch, or had a material requirement, failed.
- **B-18** — `POST`/`PATCH /libraries` accepted a `target_collection_id` that does not
  exist. Now 404 `collection_not_found` on both write paths.
- **B-19** — `vault_audit_findings.run_id` had `ondelete="CASCADE"` in the migration
  chain and never in the models, so a **fresh install refused to delete an audit run**
  and an upgraded one cascaded.

Each of B-16…B-19 returned 500 on a fresh installation and quietly succeeded with a
dangling row on an upgraded one, because the two schemas disagreed.

**Found in the frontend pass**

- **B-21 — every client-side validation message in the app was replaced by a generic
  network error.** `toast.error` takes both a thrown `ApiError` and, at twelve call
  sites, a sentence the app wrote itself ("Enter a temperature between 0 and 500.",
  "Name and folder path are required.", "Save the document before adding images."). Free
  text matches no HTTP envelope and no detail code, so the parser fell through to
  `unknown` and the user was told *"Something went wrong reaching the server. Check that
  PrintStash is running"* — because they typed a letter into a number field. A detail
  code is a single snake_case token by construction, so whitespace separates the two;
  anything with a space in it is now shown as written.
- **B-22 — seventy statements of unreachable code in the upload dialog.** `filesManifest`
  and `collectionManifest` were initialised to `null` and never set to anything but
  `null` again. Two import handlers, their branches in submit, both `SelectionList`
  renders and the component itself were dead — remnants of the URL flow from before
  captures moved to the inbox. Removed.

**Tooling and environment**

- **B-1** greenlet missing on Apple Silicon · **B-2** `audit_logs` escaped the
  between-test wipe · **B-3** three tests were quietly using the real network ·
  **B-6** two red Playwright specs driving the wrong widget · **B-7** a stale OpenAPI
  snapshot · **B-9** two parallel backup harnesses · **B-14** a dead WebdriverIO API ·
  **B-15** Chrome 137 removed `--load-extension` *silently*, so the extension suite had
  been installing nothing at all.
- **B-20** — **every lane of `./scripts/test.sh` aborted** with
  `pytest_args[@]: unbound variable` when run with no extra arguments — exactly how
  `coverage` and `fast` are documented in AGENTS.md. bash 3.2 (macOS's default shell)
  rejects `"${empty[@]}"` under `set -u`; CI runs bash 5, where it is legal, so it never
  showed up there.
- jsdom implements pointer *events* but not pointer *capture*, and dnd-kit's sensor
  calls it unconditionally — the vault suite failed its run on an unhandled error that
  pointed at no test in particular. Polyfilled in the vitest setup.

## Two things that were not about tests at all

**The suite had foreign keys switched off.** `PRAGMA foreign_keys` is ignored inside a
transaction, and the between-test wipe wrapped both pragmas in `begin()` — so it turned
enforcement off and never turned it back on, for the rest of the worker's session rather
than just the wipe. Production runs with `foreign_keys=ON`. Turning it on in the suite
surfaced B-16…B-19 immediately. There is now a parity test that asserts both the pragma
*and* that a dangling FK is actually refused, because a pragma reading 1 on one
connection proves nothing about the connection a test runs on.

**The two supported schemas had drifted 136 structural differences apart.** A fresh
install builds its schema with `create_all`; an upgrade replays the migration chain. They
were not the same database. The cause was `if not is_sqlite` guards in hand-written
migrations — SQLite cannot `ALTER` a constraint, so constraint work was simply skipped
there rather than done through `op.batch_alter_table`. Convergence is now **136 → 0**,
and the orphan-rescue path works on an upgraded database for the first time.

Guards so it cannot recur: hand-written migrations are banned (a new `database` skill
covers the workflow end to end), dialect-guarded constraint ops fail a repo test,
autogenerate refuses a possible rename rather than emitting a silent drop-and-add, empty
migrations are suppressed, and the chain round-trips head → base → head with every
migration rendering offline on both dialects.

## Structural work

- **Tiers are directories** — `unit/`, `integration/`, `contract/`, `e2e/`, `repo/` —
  and `tests/` mirrors `app/`, so "is this module tested?" is one `ls`.
- **The mirror is enforced in both suites.** Seventeen frontend files and three backend
  ones were named for a *topic* (`send-to-queue`, `small-clients`,
  `test_storage_ownership_quarantine`) while defending a real module nobody could find
  them from. All renamed or split into folders named for their module.
- **`frontend/tests/repo/`** now exists as the analogue of `backend/tests/repo/`. Adding
  `tests` to the frontend `tsconfig` typechecks the Playwright specs for the first time,
  which immediately surfaced a real error in `e2e-real/documents.spec.ts`.
- **`src/test-support/render.tsx`** is the frontend's answer to `conftest.py`: one
  `renderApp` that wires the real router, query hooks, api client, i18n catalog, auth
  context and `Toaster`, and stubs only `fetch`. It is what made the app's component
  layer testable at all — three details in it are load-bearing and each one had been
  silently breaking tests before it existed: it uses the application's own
  `queryClient` (`request.ts` invalidates *that* one, so a private client never sees the
  refresh a mutation triggers), it calls `storeLogin` as well as providing the context
  (`useRequireAuth` reads storage, so a context-only render looks signed out and hides
  every write affordance), and it mounts the real `Toaster` (most write paths report
  their outcome only through a toast).
- **Row builders**: every backend test builds its rows through `tests/factories/`;
  `_user` existed thirteen times and defaulted `superuser` to `True` in one file and
  `False` in five, so the identical call meant opposite things depending on which file
  you were reading. The frontend has the same in `src/test-support/factories.ts`.
- **Testcontainers are the only path** to PostgreSQL and S3. No Docker *fails*; it never
  skips. A skipped test is not a passing test.
- **`@printstash/ui` went from one smoke test to a mirrored test per module** — the two
  overlay hooks every dialog composes are now fully covered, including the exit-window
  and focus-restoration behaviour that no screenshot test can see.

## Tests that were green while proving nothing

Worth calling out separately, because a suite full of these is worse than a smaller
honest one:

- `test_enqueue_respects_printer_scope` asserted `0` then `1` for the **same** printer
  id — it only worked because neither id was a real row.
- `test_dispatch_claimed_raises_when_printer_missing` hard-deleted a printer out from
  under a queued job, which no code path can do.
- ~70 fixture rows referenced ids that could not exist (`owner_user_id=7`,
  `printer_id=1`, `model_id=42`).
- Three tests counted commits from a code path that never executed.
- A memory test measured RSS in a long-lived xdist worker: 393 MB one run, green the
  next. Replaced with a deterministic reachability assertion.
- The settings panel's create-a-user test had a conditional early return, so it passed
  whether or not the form was on the page. Replaced with one that fills the real fields
  and reads the POST body.

## The frontend coverage push

The `app` number went **36.44 % → 80.96 % statements, 30.76 % → 73.43 % branches**, over
~35 commits, with every floor ratcheted as it moved. Roughly forty modules went from
untested to covered: the login page, the app shell and setup gate, the auth banner and
sidebar, the vault grid and its filter sidebar, the model page and every tab under it,
the upload dialog and its whole ingestion pipeline, the settings panel's thirteen
sections, printer detail, shared volumes, notification channels, the maintenance audit,
Spoolman, storage configuration, SSO, the thumbnail cache and the session provider.

The tests are behavioural rather than structural — each one names an observable outcome
and its header says what goes wrong when it goes red. A representative sample of what
they pin, all of which was previously unprotected:

- a secret that comes back masked never travels back as the mask (printer credentials,
  Spoolman, S3, SSO, notification channels — five separate places with the same rule);
- a mesh and a G-code uploaded together are two jobs, and the second carries the first's
  content hash, or the revision lands as a separate model with no mesh beside it;
- an optimistic star reverts when the server refuses, rather than staying lit over a
  favourite that was never saved;
- a member reaching `/printers` by bookmark is put back on the vault — hiding the nav
  link is not the boundary;
- "Delete" on a model says *trash*, and restore/purge/e-stop each stop at a
  confirmation;
- a partly-failed batch names what it skipped instead of counting only the wins, and
  move and tag both offer an undo that restores each model's *original* value;
- a share link scoped to selected revisions produces a different request from one
  covering all of them.

**Where it stands honestly.** 80.96 % is not the backend's 93.62 %, and the remaining
gap is not all reachable from vitest. 1 630 statements are uncovered; ~210 of them are in
`stl-viewer` (three.js/WebGL), `gcode-viewer` (canvas), `pdf-viewer` (pdf.js worker) and
`router.tsx` (a lazy route table) — none of which renders in jsdom, and all of which are
exercised by Playwright, which v8 cannot see. The ceiling for this measurement is
therefore ~97.6 %, and the ~1 400 statements between here and there are ordinary
component work: mostly pointer-drag paths, deep branches of the four largest components,
and the long tail of small panels. It is a continuing grind rather than a blocked one,
and every point of it is now gated so it cannot slide back.

## Still open

- **F-6** — the route-level upload caps sit behind the app-wide body limit, so the
  route-level ones are unreachable. A design question, not a bug.
- The app coverage gap above.

Everything else previously listed as open is now done: F-5/F-7/F-9 (dead defensive
branches, removed), F-10 (the five identical "Edit" buttons on the Source tab, now
`aria-label`led per field), F-11 (orphaned i18n keys, removed), F-12 (the mirror rule,
enforced in both suites), the `printstash-core` URL/manifest duplication (split, with
seven behaviours carried across after coverage showed a 0.05 % drop), and the
`UniqueConstraint` naming audit (all named; now asserted rather than assumed).

---

Verified locally on this commit: backend **6 035 passed / 0 failed / 0 skipped**,
93.62 % branches, coverage floors green; `printstash-core` **1 396 passed**, 98.84 %;
frontend **1 447 passed** with `lint`, `format:check`, `typecheck` and the three-suite
coverage gate all green; Playwright mock-API **32/32**.

PR #145, opened 2026-09-07
feat(storage): authorize native artifact downloads and private HTTP caching

### LINKED ISSUE #101: [Feature]: Offload Artifact delivery with capability-safe redirects and HTTP caching
> [Download the implementation plan (`.md`)](https://github.com/user-attachments/files/31663922/printstash-native-asset-delivery-implementation-plan.md)

## Problem

PrintStash has the beginnings of provider-native download delivery, but it is
not used by the canonical product path.

On `0.13.0`, `GET /api/v1/files/{id}/download` returns a local
`FileResponse` or proxies remote bytes through FastAPI with
`StreamingResponse`. S3 presigning is exposed separately through
`download-url` and `download-direct`, and the frontend does not use either
route. Public-share downloads, thumbnails, and remote Library Source Artifacts
also pass through the API server.

This means an installation backed by S3, Cloudflare R2, Backblaze B2, Wasabi,
or another compatible object store pays API CPU, memory, connection, and
bandwidth overhead even when the provider can safely deliver the exact object
directly.

Relevant current code:

- [Canonical download and streaming helpers](https://github.com/xiao-villamor/PrintStash/blob/0.13.0/backend/app/api/v1/files.py#L80-L169)
- [Separate S3 URL/redirect routes](https://github.com/xiao-villamor/PrintStash/blob/0.13.0/backend/app/api/v1/files.py#L277-L313)
- [Public-share streaming](https://github.com/xiao-villamor/PrintStash/blob/0.13.0/backend/app/api/v1/share.py#L42-L126)
- [Artifact content integrity checks](https://github.com/xiao-villamor/PrintStash/blob/0.13.0/backend/app/services/artifact_content.py#L38-L191)
- [S3 presigning implementation](https://github.com/xiao-villamor/PrintStash/blob/0.13.0/backend/app/services/storage_backend.py#L2634-L2644)

## Proposed solution

Create one artifact-delivery module behind the authorized route surface. It
selects a structured delivery plan:

1. **Temporary redirect** when the adapter can produce a browser-safe,
   short-lived URL for the exact Artifact.
2. **Local file response** when the Artifact has a safe filesystem path.
3. **Proxy stream** as the universal fallback.

`FileResponse` is deliberately preferable to a generic stream for local
files. It populates file metadata and validators, natively handles byte ranges
and `If-Range`, and can use the ASGI server's optimized file-send path where
available. PrintStash should still evaluate `If-None-Match` and
`If-Modified-Since` in the shared delivery layer before constructing the
response. A generic `StreamingResponse` would require PrintStash to recreate
all file/range behavior and would usually move the bytes through Python-level
iteration unnecessarily.

The canonical `GET /api/v1/files/{id}/download` route should perform
authentication/RBAC and then use this module. Remove `download-url` and
`download-direct`; this API has no backward-compatibility requirement.

A redirectable URL is a narrower capability than generic "presign": it must
be HTTPS, GET-only, scoped to one object, expire quickly, carry all required
authorization in the URL, require no extra headers/cookies, preserve the
download filename, and bind to the expected immutable object identity where
the source is externally owned.

Use a temporary redirect (`307`, not `301`) with private/no-store and
no-referrer policy. A signed destination URL is a bearer credential and must
never be persisted or logged.

## Storage adapter delivery capability

Keep delivery strategy in the Artifact-delivery service, while each storage
adapter may return a structured browser-safe download target for an object it
owns. There is no pluggable edge/CDN delivery adapter in this scope.

R2 uses only the existing S3-compatible mechanism. PrintStash returns a `307`
whose `Location` is the short-lived presigned
`*.r2.cloudflarestorage.com` URL. This directly offloads the object body from
PrintStash, requires no custom domain, and is not Cloudflare CDN caching:
[R2 presigned URLs](https://developers.cloudflare.com/r2/api/s3/presigned-urls/).

R2 custom domains, Workers, WAF signing, and a configurable CDN-delivery
adapter are explicit non-goals. Do not add configuration or an opt-in path for
them.

## Conditional requests and safe browser caching

All API-served Artifacts should emit correct validators and cache policy.
`If-None-Match` with the Artifact's immutable SHA-256 is the primary validator;
`If-Modified-Since` is supported as the secondary validator, with normal HTTP
precedence rules. Range requests and `If-Range` should work wherever the local
file or remote transport supports them.

Caching must remain authorization-aware:

- authenticated originals and thumbnails: `Cache-Control: private, max-age=0,
  must-revalidate`, so the browser may retain bytes but every reuse rechecks
  PrintStash authorization and can return `304`;
- public-share/token responses: `private, no-store` by default, because share
  revocation must not be bypassed by a fresh browser cache entry;
- temporary redirects: `private, no-store`, unless a future explicitly bounded
  redirect-cache policy accepts authorization revocation latency;
- immutable long-lived caching is not introduced by this work.

The existing authenticated thumbnail response currently says `public`; this
must be corrected because a shared cache must not reuse an authenticated
Artifact response for another requester.

## Provider expectations

| Provider | Expected delivery |
| --- | --- |
| AWS S3, R2, B2, Wasabi, MinIO and compatible S3 | Presigned GET redirect |
| Local filesystem | `FileResponse` through PrintStash |
| WebDAV / Nextcloud | Proxy unless a provider-specific URL meets the browser-safe contract; never expose Basic credentials |
| SFTP | Proxy |
| Azure Blob, if added | Read-only SAS redirect |
| GCS, if added | V4 signed URL when signing credentials are available |
| OneDrive, if added | Short-lived preauthenticated `downloadUrl` |
| Dropbox, if added | Provider-specific temporary link |
| Google Drive | Proxy by default because private downloads normally need an authorization header |

OpenDAL's generic presigned request may include required headers, so successful
provider presigning alone must not imply redirect safety:
[OpenDAL PresignedRequest](https://opendal.apache.org/docs/rust/opendal/struct.PresignedRequest.html).

## Integrity and security constraints

- Authorization happens before generating a destination URL.
- Managed Vault objects can redirect only after their managed-key checks pass.
- Remote Library Source Artifacts currently copy and verify size plus SHA-256
  before delivery. Direct delivery must remain disabled unless the source can
  sign an immutable recorded version/generation; redirecting a mutable key
  would bypass the exact-content invariant.
- Public-share redirect TTL bounds revocation. Keep it very short or retain
  proxying where immediate revocation is required.
- Browser-fetch consumers such as mesh/toolpath previews need compatible CORS;
  a navigation-safe URL is not automatically fetch-safe.
- Derived/transformed responses such as 3MF/OBJ-to-STL conversion and embedded
  G-code stay inside PrintStash.

## Acceptance criteria

- [ ] One canonical authenticated file-download route selects redirect, local
      file, or proxy delivery without route-level provider checks.
- [ ] `download-url` and `download-direct` are removed from routes, frontend,
      OpenAPI fixtures, and tests.
- [ ] S3-compatible managed Vault downloads return a temporary redirect and do
      not stream the object through the API process.
- [ ] Local, WebDAV, and SFTP downloads retain a safe fallback.
- [ ] The redirect contract is structured and rejects URLs requiring headers,
      durable/public access, or an unpinned external object.
- [ ] Public share, slicer, thumbnail, original-file, and preview behavior is
      explicitly classified by delivery purpose and tested.
- [ ] Redirect responses use temporary status, short TTL, `private, no-store`,
      and `Referrer-Policy: no-referrer`.
- [ ] API-served local and proxied Artifacts implement ETag,
      `If-None-Match`, `Last-Modified`, `If-Modified-Since`, Range, and
      `If-Range` according to transport capability and HTTP precedence.
- [ ] Authenticated responses use private revalidation rather than public
      caching; share/token responses fail closed with no-store unless a
      separately reviewed policy permits bounded caching.
- [ ] Signed URLs and query credentials are redacted from logs and never stored.
- [ ] Range behavior is covered for local, redirected, and proxied responses.
- [ ] Unit/integration coverage plus one end-to-end test proves that a remote
      managed Artifact downloads without its bytes crossing the API server.
- [ ] Operator docs explain that R2 uses its S3 API presigned URL and that this
      work does not support a custom-domain/CDN delivery mode.

## Alternatives considered

- **Keep the separate direct-download route:** rejected because normal product
  links continue to choose the slower path and provider logic leaks into the
  API surface.
- **Always return a URL in JSON:** rejected because clients must duplicate
  delivery selection and URL-bearing responses are easier to persist or log.
- **Make object-store paths public:** rejected; delivery must use short-lived,
  object-scoped authorization.
- **Redirect every remote source:** rejected because WebDAV/SFTP cannot provide
  browser-safe URLs and mutable Library Sources require PrintStash verification.

xiao-villamor:
# Artifact delivery validation

The acceptance scenarios below cover issue #101. Focused results: 61 API/unit
cases, 55 provider/unit/E2E cases, 17 frontend request cases and the reviewed
OpenAPI contract passed. The real Chromium native-download proof passed one
test (19.7 seconds overall), including exact saved bytes, Unicode filename,
provider CORS and zero API body bytes. The complete measured backend run passed
8,768 non-resource tests and 153 real-resource tests. One deterministic staging
cleanup case was then appended to that report to remove an order-sensitive
dependency pin: aggregate coverage is 93.96%, `artifact_delivery.py` is 90.39%,
and all 10 coverage-floor checks pass.

| # | Behaviour (test name) | Category | Precondition / input | Observable outcome asserted | Tier | Status |
|---|---|---|---|---|---|---|
| 1 | revalidates_authenticated_original_privately | Happy | Authenticated original | Private revalidation header | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_revalidates_authenticated_original_privately` |
| 2 | uses_the_original_digest_as_validator | Happy | Original bytes | ETag equals original SHA-256 | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_uses_the_original_digest_as_validator` |
| 3 | returns_no_body_for_matching_validator | Edge | Matching weak ETag | 304 with empty body | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_returns_no_body_for_matching_validator` |
| 4 | prioritizes_etag_over_modification_date | Edge | Different ETag, future date | Original bytes returned | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_prioritizes_etag_over_modification_date` |
| 5 | evaluates_modification_date | Edge | Unchanged date | 304 response | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_evaluates_modification_date` |
| 6 | authorizes_before_conditional_response | Error | Anonymous request, wildcard ETag | 401 before validation | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_authorizes_before_conditional_response` |
| 7 | serves_exact_local_range | Happy | Local bytes=2-7 | 206 with exact segment | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_serves_exact_local_range` |
| 8 | ignores_range_when_if_range_does_not_match | Edge | Stale If-Range | 200 with full original | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_ignores_range_when_if_range_does_not_match` |
| 9 | removes_obsolete_delivery_route | Error | Both retired route suffixes | 404 response | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_removes_obsolete_delivery_route` |
| 10 | denies_revoked_conditional_access | Error | Revoked share with matching ETag | 404 before validation | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_denies_revoked_conditional_access` |
| 11 | keeps_shared_original_noncacheable | Happy | Public share download | private, no-store | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_keeps_shared_original_noncacheable` |
| 12 | keeps_slicer_original_noncacheable | Happy | Slicer capability download | private, no-store | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_keeps_slicer_original_noncacheable` |
| 13 | revalidates_authenticated_thumbnail_privately | Happy | Authenticated thumbnail | Private revalidation header | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_revalidates_authenticated_thumbnail_privately` |
| 14 | uses_a_separate_converted_validator | Edge | OBJ converted to STL | Different representation ETag | Integration | ✅ `integration/api/v1/files/test_delivery.py::TestAuthorizedDelivery::test_uses_a_separate_converted_validator` |
| 15 | offloads_the_original | Happy | HTTPS S3 with CORS | Exact original bytes from signed GET | Contract | ✅ `contract/modules/storage/test_artifact_delivery.py::TestBrowserDelivery::test_offloads_the_original` |
| 16 | preserves_the_download_filename | Happy | Unicode display name | RFC 5987 response filename | Contract | ✅ `contract/modules/storage/test_artifact_delivery.py::TestBrowserDelivery::test_preserves_the_download_filename` |
| 17 | falls_back_without_cors | Error | No bucket CORS policy | No browser target | Contract | ✅ `contract/modules/storage/test_artifact_delivery.py::TestBrowserDelivery::test_falls_back_without_cors` |
| 18 | serves_selected_range | Happy | Remote original, Range | 206 with exact segment | Contract | ✅ `contract/modules/storage/test_artifact_delivery.py::TestBrowserDelivery::test_serves_selected_range` |
| 19 | compares_if_range_to_the_original_digest | Edge | Range with matching SHA validator | 206 despite distinct provider ETag | Contract | ✅ `contract/modules/storage/test_artifact_delivery.py::TestBrowserDelivery::test_compares_if_range_to_the_original_digest` |
| 20 | serves_full_original_for_stale_if_range | Edge | Remote Range with stale validator | 200 full original | Contract | ✅ `contract/modules/storage/test_artifact_delivery.py::TestBrowserDelivery::test_serves_full_original_for_stale_if_range` |
| 21 | accepts_an_object_scoped_target | Happy | HTTPS, matching object/origin, 60 seconds | Target accepted | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_accepts_an_object_scoped_target` |
| 22 | refuses_an_unsafe_url | Error | HTTP, userinfo, fragment, control, relative, missing host | Target rejected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_refuses_an_unsafe_url` |
| 23 | refuses_invalid_expiry | Error | Expired, zero, over 60 seconds | Target rejected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_refuses_invalid_expiry` |
| 24 | refuses_a_different_object | Error | Object identity differs | Target rejected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_refuses_a_different_object` |
| 25 | refuses_unproven_cors | Error | Browser origin lacks proof | Target rejected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_refuses_unproven_cors` |
| 26 | refuses_required_headers | Error | Target needs extra authorization header | Target rejected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_refuses_required_headers` |
| 27 | redacts_target_representation | Error | Secret query and object key | Neither exposed by repr | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_redacts_target_representation` |
| 28 | matches_an_original_validator | Edge | Weak ETag, list, wildcard | Conditional match | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestNotModified::test_matches_an_original_validator` |
| 29 | ignores_invalid_date | Error | Malformed HTTP date | Full response selected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestNotModified::test_ignores_invalid_date` |
| 30 | selects_the_requested_bytes | Edge | Closed, open, suffix, clipped ranges | Expected inclusive bounds | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestByteRange::test_selects_the_requested_bytes` |
| 31 | rejects_unsatisfiable_range | Error | Reversed, past end, zero suffix, oversized integer | Unsatisfiable error | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestByteRange::test_rejects_unsatisfiable_range` |
| 32 | ignores_unsupported_range | Edge | Malformed or multipart range | Full response selected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestByteRange::test_ignores_unsupported_range` |
| 33 | weak_validator_cannot_select_a_partial_response | Edge | Weak If-Range | Full response selected | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestRangeMatches::test_weak_validator_cannot_select_a_partial_response` |
| 34 | matches_modification_time | Edge | Exact modification date | Range allowed | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestRangeMatches::test_matches_modification_time` |
| 35 | encodes_unicode_display_name | Happy | Non-ASCII filename | UTF-8 filename parameter | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestContentDisposition::test_encodes_unicode_display_name` |
| 36 | removes_path_header_injection | Error | Traversal and CRLF filename | Safe display name | Unit | ✅ `unit/modules/storage/test_artifact_delivery.py::TestContentDisposition::test_removes_path_header_injection` |
| 37 | falls_back_when_cors_policy_is_unreadable | Error | Denied GetBucketCors | API fallback | Unit | ✅ `unit/modules/storage/storage_backend/test_s3.py::TestBrowserDownload::test_falls_back_when_cors_policy_is_unreadable` |
| 38 | refuses_a_later_permissive_cors_rule | Error | Earlier matching rule does not expose filename | API fallback | Unit | ✅ `unit/modules/storage/storage_backend/test_s3.py::TestBrowserDownload::test_refuses_a_later_permissive_cors_rule` |
| 39 | falls_back_when_the_signer_fails | Error | Signer exception with sensitive diagnostics | Fallback without secret logging | Unit | ✅ `unit/modules/storage/storage_backend/test_s3.py::TestBrowserDownload::test_falls_back_when_the_signer_fails` |
| 40 | closes_an_unstarted_body | Error | Consumer closes before first read | Provider body closed | Unit | ✅ `unit/modules/storage/storage_backend/test_s3.py::TestStreamRange::test_closes_an_unstarted_body` |
| 41 | rejects_a_truncated_body | Error | Provider body shorter than advertised | Operation error and closed body | Unit | ✅ `unit/modules/storage/storage_backend/test_s3.py::TestStreamRange::test_rejects_a_truncated_body` |
| 42 | offloads_authorized_download | Happy | Canonical authenticated API + HTTPS S3 | 307 with empty API body; exact provider bytes | E2E | ✅ `e2e/test_artifact_delivery.py::TestArtifactDelivery::test_offloads_authorized_download` |
| 43 | retries a failed browser delivery through the API | Error | Fetch raises TypeError | One retry using original API and proxy header | Frontend unit | ✅ `src/lib/api/__tests__/request.test.ts::retries a failed browser delivery through the API` |
| 44 | does not retry a cancelled read | Edge | Aborted signal | Cancellation propagated | Frontend unit | ✅ `src/lib/api/__tests__/request.test.ts::does not retry a cancelled read` |
| 45 | does not retry an authorization denial | Error | 403 API response | Denial propagated once | Frontend unit | ✅ `src/lib/api/__tests__/request.test.ts::does not retry an authorization denial` |
| 46 | preserves cancellation while revalidating protected text | Edge | Text read with signal | Signal and no-cache mode retained | Frontend unit | ✅ `src/lib/api/__tests__/request.test.ts::preserves cancellation while revalidating protected text` |
| 47 | downloads S3 bytes through the authenticated helper | Happy | Actual browser, authenticated API, HTTPS S3 with CORS, Unicode name | Exact bytes and Unicode filename; no provider app credentials or referrer; sole API response is private/no-store 307 with zero emitted body bytes | Playwright | ✅ `frontend/tests/e2e-real/delivery/native-download.spec.ts::native artifact delivery::downloads S3 bytes through the authenticated helper` |

The browser run used the actual API/TLS provider fixture started once before a
coordinated stable-network window, then reused it through the same delivery
config. An isolated offline dependency installation resolved a changing shared
Playwright installation before the passing run. Backend fixture lint, launcher
syntax validation, frontend test lint and TypeScript all passed.

## Attached-plan observability and contract follow-up

80 focused delivery/thumbnail/telemetry tests and 290 adapter regressions
passed. The completed backend measurement also passed all 8,768 non-resource
tests and all 153 real-resource tests, including the additional inline-provider
contract. The follow-up raised delivery-module coverage from the earlier 88.57%
to 90.39% and the final 10-test floor audit passed.

| # | Behaviour (test name) | Category | Precondition / input | Observable outcome asserted | Tier | Status |
|---|---|---|---|---|---|---|
| 48 | records selected delivery strategy | Happy | Authorized local download | Counter increases once | Integration | ✅ `backend/tests/integration/api/v1/files/test_delivery.py::TestDeliveryObservability::test_records_selected_delivery_strategy` |
| 49 | counts consumed proxy bytes | Happy | Partially consumed stream | Only consumed byte count charged | Unit | ✅ `backend/tests/unit/modules/storage/test_delivery_observability.py::TestDeliveryObservability::test_counts_consumed_proxy_bytes` |
| 50 | redacts unknown delivery labels | Error | URL-shaped telemetry labels | No bearer credential in exposition | Unit | ✅ `backend/tests/unit/modules/storage/test_delivery_observability.py::TestDeliveryObservability::test_redacts_unknown_delivery_labels` |
| 51 | releases metered stream on disconnect | Error | Early and repeated close | Underlying stream closes once | Unit | ✅ `backend/tests/unit/modules/storage/test_delivery_observability.py::TestDeliveryObservability::test_releases_metered_stream_on_disconnect` |
| 52 | reports delivery capability without signing | Happy | Local detailed health | Safe delivery capability object | Integration | ✅ `backend/tests/integration/api/v1/files/test_delivery.py::TestDeliveryObservability::test_reports_delivery_capability_without_signing` |
| 53 | redacts signed queries in log records | Error | Message/access query credentials | Formatted output is redacted | Unit | ✅ `backend/tests/unit/core/test_logging.py::TestSensitiveQueryFilter::test_redacts_signed_queries_in_log_records` |
| 54 | redacts exception query credentials | Error | SDK traceback contains signedURL | Formatted exception excludes credential | Unit | ✅ `backend/tests/unit/core/test_logging.py::TestSensitiveQueryFilter::test_redacts_exception_query_credentials` |
| 55 | preserves Uvicorn access formatter arguments | Edge | Structured access log tuple | Valid redacted request line | Unit | ✅ `backend/tests/unit/core/test_logging.py::TestSensitiveQueryFilter::test_preserves_uvicorn_access_formatter_arguments` |
| 56 | preserves nonsensitive mapping log arguments | Edge | Mapping with numeric status | Formatting preserves status and scrubs query | Unit | ✅ `backend/tests/unit/core/test_logging.py::TestSensitiveQueryFilter::test_preserves_nonsensitive_mapping_log_arguments` |
| 57 | closed proxy stream stays closed | Edge | Read after close | No further bytes | Unit | ✅ `backend/tests/unit/modules/storage/test_delivery_observability.py::TestDeliveryObservability::test_closed_proxy_stream_stays_closed` |
| 58 | propagates provider stream failure | Error | Provider raises mid-read | Error retained and stream closed | Unit | ✅ `backend/tests/unit/modules/storage/test_delivery_observability.py::TestDeliveryObservability::test_propagates_provider_stream_failure` |
| 59 | keeps proxy bytes when metrics fail | Error | Metric backend fails | Original bytes preserved | Unit | ✅ `backend/tests/unit/modules/storage/test_delivery_observability.py::TestDeliveryObservability::test_keeps_proxy_bytes_when_metrics_fail` |
| 60 | keeps delivery when strategy metrics fail | Error | Metric backend fails | Selection continues | Unit | ✅ `backend/tests/unit/modules/storage/test_delivery_observability.py::TestDeliveryObservability::test_keeps_delivery_when_strategy_metrics_fail` |
|61|removes bare URL contract|Edge|Public content/backend interfaces|Only structured targets exposed|Unit|✅ `test_artifact_delivery.py::TestStorageDeliveryContract::test_removes_bare_url_contract`|
|62|preserves inline thumbnail disposition|Happy|Authenticated thumbnail|Inline filename response|Integration|✅ `test_delivery.py::TestAuthorizedDelivery::test_preserves_inline_thumbnail_disposition`|
|63|preserves inline provider disposition|Happy|Real S3 thumbnail target|Exact inline filename/media type|Contract|✅ `test_artifact_delivery.py::TestBrowserDelivery::test_preserves_inline_provider_disposition`|
|64|refuses unparseable browser URL|Error|Malformed IPv6 authority from provider|Capability rejected without escaping validation|Unit|✅ `backend/tests/unit/modules/storage/test_artifact_delivery.py::TestSafeBrowserDownload::test_refuses_unparseable_url`|
|65|quarantines owned staging file without xattrs|Edge|Owned file; filesystem reports `ENOTSUP` for marker reads|Inode proof permits exact-file cleanup and coverage is order-independent|Integration|✅ `backend/tests/integration/modules/ingestion/test_staging_leases.py::TestEntryHelpers::test_quarantines_owned_file_without_xattrs`|

Closes #101


docs/artifact-delivery.md
```diff
@@ -0,0 +1,66 @@
...
+exposure. Redirect URLs are bearer credentials; they are not persisted.
+
+Authenticated originals and thumbnails use `private, max-age=0, must-revalidate`.
+Shares and slicer responses use `private, no-store`. Original validators use
+SHA-256; converted content uses its own representation validator. ETag conditions
+take precedence over modification dates. Requests with Range retain API delivery because the provider ETag can differ
+from the original SHA-256. S3 proxy reads support one byte range;
+unsupported multi-range syntax falls back to the full representation.
+
+## Browser configuration
```
jorgehermo9, line 22: BTW @xiao-villamor , remember that this is needed in the bucket configuration I think

xiao-villamor: Correct — the bucket owner must configure this. I added the exact S3 CORS JSON under Browser configuration and clarified that PrintStash only validates the first matching rule; it never creates or modifies bucket CORS. Without a readable compatible rule, delivery safely falls back to the API proxy. The docs also now require the storage endpoint to use a different origin so the PrintStash Authorization header cannot be retained across the 307.

xiao-villamor:
## Coverage matrix addendum after the final plan audit

The original PR matrix remains valid. The literal pass against the attached issue plan added the following explicit mappings; the canonical 79-row matrix is in `docs/artifact-delivery-validation.md`.

| # | Behaviour | Category | Observable outcome | Tier | Status |
|---|---|---|---|---|---|
|66|Public-share download capability|Error|View-only share is denied before delivery|Integration|✅|
|67|Expired public share|Error|Expired token is rejected before delivery|Integration|✅|
|68|Public-share file scope|Error|Another model's file remains hidden|Integration|✅|
|69|Slicer artifact scope|Error|Token for another Artifact is rejected|Integration|✅|
|70|Missing managed blob|Error|Returns `410 file_blob_missing`|Integration|✅|
|71|Remote Vault proxy|Happy|Streaming-only backend returns exact bytes through API|Integration|✅|
|72|WebDAV/SFTP native-delivery exclusion|Edge|OpenDAL exposes no browser target|Unit|✅|
|73|Remote Library Source verification|Happy|Catalog-bound bytes are materialized without Vault access|Integration|✅|
|74|Changed remote source|Error|Integrity change fails before exposure|Integration|✅|
|75|Mounted source access|Happy|Stream/materialize preserve digest-pinned bytes|Integration|✅|
|76|Changed mounted source|Error|Stale catalog digest fails before exposure|Integration|✅|
|77|Canonical OpenAPI|Edge|Only the canonical route surface remains|Repo contract|✅|
|78|STL passthrough/conversion|Happy|Original STL passes through; other meshes are converted|Integration|✅|
|79|Same-origin provider target|Error|API proxy prevents forwarding PrintStash Authorization to storage|Unit|✅|

Security diff scan `cad90af5..091c3e34`: 21/21 source files reviewed, complete coverage, zero reportable findings. The scan specifically covered signed URL secrecy, authorization ordering, immutable identity, CORS, cache/range semantics, log/metric exposure, browser fallback and credential forwarding.

xiao-villamor:
Final plan/security follow-up: matrix row 80 is now in `docs/artifact-delivery-validation.md`. It maps the two Python 3.13 regressions for header-only internal request adapters; both now pass, together with all delivery security unit cases (39/39). The final immutable security scan covers `cad90af5..70b21fce`, reviewed 21/21 changed source files, and reports complete coverage with zero findings.

